Issue #38, opened 2026-05-07
fix(on-recovery): @pre.7 produces wrappedKEK entries; hub expects wrappedDeks


vLannaAi:
## Problem

`@noy-db/on-recovery@0.1.0-pre.7`'s `generateRecoveryCodeSet` produces entries with a `wrappedKEK` field:

```ts
// packages/on-recovery/src/index.ts (paraphrased from dist)
async function generateRecoveryCodeSet(opts) {
  ...
  const wrappedKEK = await wrapKEK(opts.kek, wrappingKey)
  entries.push({ codeId, salt, wrappedKEK: base64Encode(wrappedKEK), enrolledAt })
}
async function wrapKEK(kek, wrappingKey) {
  return crypto.subtle.wrapKey('raw', kek, wrappingKey, 'AES-KW')   // requires extractable KEK
}
```

But the hub's `PaperRecoveryEntry` (`packages/hub/src/team/recovery.ts:43`) expects:

```ts
interface PaperRecoveryEntry {
  readonly codeId: string
  readonly salt: string
  readonly iv: string                // AES-GCM IV — on-recovery doesn't produce this
  readonly wrappedDeks: string       // AES-GCM ciphertext of { deks: { collection: rawDekBase64 } }
  readonly enrolledAt: string
}
```

The hub's design comment at `team/recovery.ts:28-42` explains: hub's KEK is non-extractable, so recovery wraps DEKs (serialized + AES-GCM-encrypted with the code-derived key), not the KEK. Same pattern as on-pin's `wrappedKeyring`.

## Two consequences

1. **on-recovery's output is structurally incompatible with `db.enrollRecovery({ profile: 'paper', entries })`** — the hub stores them but `unwrapDeksFromPaperEntry` (used by `db.recoverPassphrase`) expects `entry.wrappedDeks` and `entry.iv`, which don't exist on on-recovery's output.

2. **on-recovery only works against an extractable KEK** — but the hub deliberately builds non-extractable KEKs. So even the `wrappedKEK` it produces couldn't be unwrapped by anything in the hub's recovery flow.

## Proposed fix

Two options:

### Option A: Rewrite `@noy-db/on-recovery` to call hub helpers
Make `generateRecoveryCodeSet` delegate to `mintPaperRecoveryEntry` + handle code generation/parsing only:

```ts
import { mintPaperRecoveryEntry } from '@noy-db/hub'

async function generateRecoveryCodeSet({ deks, count }) {
[...]

Issue #26, opened 2026-05-07
feat(hub): db.enrollPassword(ceremony) parallel to db.enrollWebAuthn


vLannaAi:
## Problem

`@noy-db/on-password`'s `enrollPasswordAuthenticator(keyring, options)` calls `crypto.subtle.wrapKey('raw', keyring.kek, ...)` to wrap the KEK under the password-derived key. `wrapKey` requires the KEK to be **extractable**. Hub's `deriveKey` (in `loadKeyring`) builds the KEK with `extractable: false` (correct for normal operation, since DEK wrapping doesn't need export).

Result: `enrollPasswordAuthenticator` cannot be called against a hub-derived keyring. **Tier-2 daily-password (the whole reason `@noy-db/on-password` exists) is unreachable from a real consumer.**

## Repro

```ts
import { createNoydb } from '@noy-db/hub'
import { enrollPasswordAuthenticator } from '@noy-db/on-password'

const db = await createNoydb({ store, user: 'alice', secret: 'phrase' })
await db.openVault('niwat')
const keyring = await /* hub-internal */ db.getKeyring('niwat')
await enrollPasswordAuthenticator(keyring, { password: 'daily-pass' })
// → InvalidAccessException: key is not extractable
```

Stack:
```
node_modules/@noy-db/on-password/src/index.ts:73    subtle.wrapKey('raw', keyring.kek, wrappingKey, 'AES-KW')
```

## Proposed fix

Parallel to `db.enrollWebAuthn(vault, ceremony)` which the hub already provides for the same shape of problem (#16):

```ts
await db.enrollPassword('niwat', async (keyring) => {
  return enrollPasswordAuthenticator(keyring, { id: 'password-daily', password })
})
```

Hub provides the ceremony with a freshly-derived **extractable** keyring, runs the on-password call, persists the slot via the existing `enrollAuthenticator` path. Same machinery as `db.enrollWebAuthn`.

## Why high priority

Blocks Niwat (first production consumer). Tier-2b daily-password is the only non-biometric login path for shared devices (office desktop, RDP'd WorkSpace) — without it, those contexts have no choice but to type the rarely-typed master phrase, which defeats the LUKS multi-slot model `@noy-db/on-password` exists to enable.

Tracked downstream at: niwat#34 phase 2 (deferred).

vLannaAi:
[...]

Issue #35, opened 2026-05-07
feat(hub): export buildUnlockedKeyringFromKek


vLannaAi:
## Problem

\`@noy-db/on-password\`'s \`verifyPasswordSlot\` requires a \`materialize\` callback:

```ts
verifyPasswordSlot(slot, password, options: VerifyPasswordSlotOptions): Promise<UnlockedKeyring>

interface VerifyPasswordSlotOptions {
  /**
   * Caller-supplied "given the recovered KEK, build the
   * UnlockedKeyring" routine. The hub provides the standard
   * implementation in {@link buildUnlockedKeyringFromKek}; consumers
   * with non-standard storage can pass their own.
   */
  readonly materialize: (kek: CryptoKey) => Promise<UnlockedKeyring>
}
```

The docstring promises \`buildUnlockedKeyringFromKek\` as the canonical implementation — but it's **not in \`@noy-db/hub\`'s public exports** in pre.7. Consumers must inline the equivalent (~25 lines: load \`_keyring/<userId>\` envelope, parse the \`KeyringFile\`, AES-KW-unwrap each DEK with the supplied KEK, build the \`UnlockedKeyring\` object).

Niwat ships this inline in \`niwat-app/app/auth/pre-auth.ts\` as \`materializeFromKek\`. The implementation duplicates hub's internal \`loadKeyring\` (in \`@noy-db/hub/src/team/keyring.ts\`) minus the passphrase-derivation step.

## Proposed fix

Export \`buildUnlockedKeyringFromKek\` from \`@noy-db/hub\`:

```ts
/**
 * Build an UnlockedKeyring from a recovered KEK + the user's keyring file.
 * Used by tier-2 ceremonies that recover the KEK directly (on-password,
 * on-oidc) — passed as the `materialize` callback.
 */
export async function buildUnlockedKeyringFromKek(
  store: NoydbStore,
  vault: string,
  userId: string,
  kek: CryptoKey,
): Promise<UnlockedKeyring>
```

The implementation already exists internally — this is just an export.

## Why low priority

Workaround is trivial (~25 LOC). But it's a docstring contract that doesn't match runtime exports — slightly confusing for new consumers, easy to fix.

Tracked downstream at: niwat#34 phase 2 (\`materializeFromKek\` in \`pre-auth.ts\`).

Related: #26 (\`db.enrollPassword\`), #27 (\`generateRecoveryCodeSet\` extractable-KEK).

vLannaAi:
[...]

PR #40, opened 2026-05-07
feat(hub): public surface for native paper recovery (#28, #39)

### LINKED ISSUE #28: feat(hub): public db.getKeyring(vault) accessor
## Problem

Several legitimate consumer flows need access to the live unlocked keyring after \`createNoydb\`:

- Tier-3 PIN enrolment via \`@noy-db/on-pin\`'s \`enrollPin(keyring, options)\` — needs the keyring's DEKs.
- Tier-2 enrolment ceremonies that don't go through \`db.enrollWebAuthn\` / \`db.enrollPassword\` (e.g. custom \`@noy-db/on-*\` packages).
- The \`materialize\` callback for \`verifyPasswordSlot\` (when implemented inline pending hub-side helper export — see #N for buildUnlockedKeyringFromKek).

Hub keeps the keyring cached internally but doesn't expose a public accessor. Consumers reach in via:

```ts
const keyring = await (db as unknown as {
  getKeyring(vault: string): Promise<UnlockedKeyring>
}).getKeyring(vault)
```

The internal getter is documented as \`@internal\` but is the only path. Brittle (any internal rename breaks consumers) and bypasses any access control.

## Proposed fix

Add a public method:

```ts
class Noydb {
  /**
   * Get the unlocked keyring for a vault. Gated by minTier:1 to avoid
   * leaking tier-3 keyrings to elevation-required code paths.
   * Throws if the vault is locked or the active session doesn't meet
   * the gate's tier requirement.
   */
  async getKeyring(vault: string): Promise<UnlockedKeyring>
}
```

Optional gate: callers passing a \`presented: { factors }\` proof for stricter cases.

## Why high priority

Required by every \`@noy-db/on-*\` package's enrol path that doesn't already have a hub-side wrapper. \`db.enrollWebAuthn\` solved this for WebAuthn (#16 in this repo's history); \`db.enrollPassword\` (#26) and \`db.enrollRecovery\` (#27) are pending equivalents — those issues effectively need this accessor as a building block.

Tracked downstream at: niwat#34 (multiple call sites use the \`(db as unknown as ...)\` cast).

### LINKED ISSUE #39: feat(hub): export mintPaperRecoveryEntry and unwrapDeksFromPaperEntry
## Problem

Both helpers are defined in [\`packages/hub/src/team/recovery.ts\`](https://github.com/vLannaAi/noy-db/blob/main/packages/hub/src/team/recovery.ts) (lines 136 and 173), included in the runtime bundle, and they are the canonical (working) primitives for native paper recovery — but they're **not in the package's public exports**.

\`packages/hub/src/index.ts\` only re-exports:
- \`loadPaperRecoveryEntries\`
- \`savePaperRecoveryEntries\`
- \`burnPaperRecoveryEntry\`
- \`hasRecoveryEnrolled\`

…from the recovery module. The two helpers a consumer actually needs to enroll codes are missing.

## Why this matters

Without the export, consumers can't enroll paper recovery natively in pre.7 at all:
- \`@noy-db/on-recovery\` is broken (#38 — produces incompatible \`wrappedKEK\` entries).
- The hub's mint helper exists but is unreachable.
- \`db.enrollRecovery({ profile: 'paper', entries })\` accepts entries but consumers can't construct them.

## Proposed fix

One-line export from \`@noy-db/hub\`:

\`\`\`ts
// packages/hub/src/index.ts
export {
  // existing
  loadPaperRecoveryEntries,
  savePaperRecoveryEntries,
  burnPaperRecoveryEntry,
  hasRecoveryEnrolled,
  // add:
  mintPaperRecoveryEntry,
  unwrapDeksFromPaperEntry,
  type PaperRecoveryEntry,
  type PaperRecoveryDoc,
} from './team/recovery.js'
\`\`\`

(\`PaperRecoveryEntry\` and \`PaperRecoveryDoc\` may already be exported via \`types-D-6bmD2c.d.ts\` re-exports — confirm at fix time.)

## Why high priority

This unblocks **all native paper recovery for every consumer in pre.7**. Combined with #38 (which can be deferred or just deprecate the on-recovery package), exporting these helpers means consumers have a working path today.

Niwat is shipping the helpers inlined locally (vendored ~70 LOC) until this lands; happy to PR the export if useful.

Tracked downstream: vLannaAi/niwat#35.

Related: #38 (on-recovery package is stale relative to hub).

vLannaAi:
## Summary

- **#28** — `db.getKeyring(vault)` is now public. Required by every `@noy-db/on-*` ceremony that needs the live DEK set; consumers were reaching in via `(db as unknown as { getKeyring }).getKeyring`.
- **#39** — `mintPaperRecoveryEntry` + `unwrapDeksFromPaperEntry` (+ `PaperRecoveryEntry` / `PaperRecoveryDoc` types) are now exported from `@noy-db/hub`. Niwat had been vendoring ~70 LOC locally as a workaround.
- **`db.enrollRecovery` docstring fix** — the previous example pointed at `@noy-db/on-recovery@<=pre.7`'s `generateRecoveryCodeSet`, but the formats are incompatible (tracked as #38). Updated to point at `mintPaperRecoveryEntry` from hub directly.

## Why now

Niwat reviewed the post-pre.7 auth surface and discovered the format mismatch + missing exports — native paper recovery had no working path despite `db.enrollRecovery({ profile: 'paper', entries })` shipping in pre.5. This PR is the small, self-contained unblock.

## Not in this PR

- **#26** (tier-2 wrap-DEKs `KeyringAuthenticator` variant — Niwat's Path C proposal that unifies tier-0/2/3 around the wrap-DEKs primitive)
- **#38** (rewrite `@noy-db/on-recovery` to delegate to `mintPaperRecoveryEntry`)

Both ship as PR1b — that's the architectural unification, this is the consumer-unblock.

## Test plan

- [x] `pnpm vitest run packages/hub` — 1304 tests pass (89 files)
- [x] `pnpm turbo typecheck --filter=@noy-db/hub` — clean
- [x] `pnpm turbo lint --filter=@noy-db/hub` — clean
- [x] New file `packages/hub/__tests__/pr1a-public-recovery.test.ts` pins the public-surface contract — imports from `../src/index.js` (the barrel) so removing either export breaks the test
- [ ] Niwat sanity-check: replace the vendored ~70 LOC with `import { mintPaperRecoveryEntry } from '@noy-db/hub'` and verify the full enroll → recover round-trip

Closes #28
Closes #39
References #38 (docstring fix; full resolution in PR1b)

### DIFF
--- a/packages/hub/__tests__/pr1a-public-recovery.test.ts
+++ b/packages/hub/__tests__/pr1a-public-recovery.test.ts
@@ -1,0 +1,124 @@
+/**
+ * PR1a — public-API exposure for paper recovery + db.getKeyring.
+ *
+ * Covers:
+ *   - issue #28: db.getKeyring(vault) is reachable from outside the hub
+ *     and returns the live UnlockedKeyring (same shape on-* packages
+ *     expect — .deks Map, .kek CryptoKey, .role).
+ *   - issue #39: mintPaperRecoveryEntry + unwrapDeksFromPaperEntry are
+ *     re-exported from @noy-db/hub and produce entries that survive
+ *     the full db.enrollRecovery → db.recoverPassphrase round-trip.
+ *
+ * The test imports from '../src/index.js' (the barrel) — not from
+ * 'team/recovery.js' — so it pins the public-surface contract that
+ * pre.8 ships. Removing either export would break this test.
+ */
+import { describe, it, expect, beforeEach } from 'vitest'
+import type { NoydbStore, EncryptedEnvelope } from '../src/types.js'
+import {
+  createNoydb,
+  type Noydb,
+  mintPaperRecoveryEntry,
+  unwrapDeksFromPaperEntry,
+  type PaperRecoveryEntry,
+} from '../src/index.js'
+
+function inlineMemory(): NoydbStore {
+  const store = new Map<string, Map<string, Map<string, EncryptedEnvelope>>>()
+  function gc(c: string, col: string) {
+    let comp = store.get(c)
+    if (!comp) { comp = new Map(); store.set(c, comp) }
+    let coll = comp.get(col)
+    if (!coll) { coll = new Map(); comp.set(col, coll) }
+    return coll
+  }
+  return {
+    name: 'inline-memory',
+    async get(c: string, col: string, id: string) { return gc(c, col).get(id) },
+    async put(c: string, col: string, id: string, env: EncryptedEnvelope) { gc(c, col).set(id, env) },
+    async delete(c: string, col: string, id: string) { gc(c, col).delete(id) },
+    async list(c: string, col: string) { return [...gc(c, col).keys()] },
+    async loadAll() { return {} },
+    async saveAll() {},
+    capabilities: { casAtomic: true, auth: { kind: 'none' } },
+  } as unknown as NoydbStore
+}
+
+const STRONG_OLD = 'correct horse battery staple printer toaster'
+const STRONG_NEW = 'glasses cabinet bicycle umbrella thunder velvet'
+
+describe('PR1a public surface', () => {
+  let db: Noydb
+
+  beforeEach(async () => {
+    db = await createNoydb({
+      store: inlineMemory(),
+      user: 'alice',
+      secret: STRONG_OLD,
+    })
+    await db.openVault('acme')
+  })
+
+  describe('db.getKeyring (#28)', () => {
+    it('returns the live UnlockedKeyring with non-empty DEK map', async () => {
+      const keyring = await db.getKeyring('acme')
+      expect(keyring.userId).toBe('alice')
+      expect(keyring.role).toBe('owner')
+      expect(keyring.deks.size).toBeGreaterThan(0)
+      expect(keyring.kek).toBeInstanceOf(CryptoKey)
+    })
+
+    it('returns the same instance on repeated calls (cached)', async () => {
+      const a = await db.getKeyring('acme')
+      const b = await db.getKeyring('acme')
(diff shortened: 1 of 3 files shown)


vLannaAi:
**Niwat consumer-side review** — LGTM.

This closes #28 + #39 cleanly with a self-contained test that pins the public-surface contract. The barrel-import in `pr1a-public-recovery.test.ts` (`from '../src/index.js'`, not internal paths) is the right call: removing either re-export breaks the test, so the contract pre.8 ships is locked in.

Two non-blocking suggestions:

**1. Public `db.options.store` accessor would clean up a test-code wart.** The third test reaches in via `(db as unknown as { options: { store: NoydbStore } }).options.store` to share a store across two `createNoydb` instances. A typed accessor would avoid the cast in tests and downstream consumer code (Niwat hits the same pattern). Could land separately.

**2. Document the recovery-code normalization expectation in `db.enrollRecovery`'s docstring.** The test comment says:
> Codes must be already-normalized (uppercase, no separators) so they round-trip through recoverPassphrase's normalizePaperCode step.

That's a real consumer contract — but the public docstring's example uses `'CORRECT-HORSE-1'` (with dashes). If `recoverPassphrase` strips dashes during normalization, the example works; if not, it doesn't. Worth a one-liner in the docstring: "the code passed at recovery time is normalized (uppercase, separators stripped) before matching, so mint with the unnormalized form you'll show the user." Saves consumers from rediscovering this the hard way (Niwat's #35 hit the same trap before settling on `parseRecoveryCode` to bridge the form mismatch).

Happy to retarget Niwat's vendored helpers to the hub barrel as soon as this hits npm.

Issue #41, opened 2026-05-07
feat(hub): tighten UnlockedKeyring.kek to CryptoKey | null


vLannaAi:
## Problem

\`UnlockedKeyring.kek\` is typed as \`CryptoKey\` at \`packages/hub/src/team/keyring.ts:72\`, but runtime code already passes \`null\` in 4 places via \`null as unknown as CryptoKey\`:

- \`packages/hub/src/noydb.ts:116\` — \`createPlaintextKeyring\` (unencrypted mode)
- \`packages/hub/src/session/session.ts:251\` — session restore
- \`packages/hub/src/session/dev-unlock.ts:266\` — dev unlock
- \`packages/on-pin/src/index.ts:349\` — tier-3 PIN resume

The type lies about what \`kek\` can be at runtime. With #26 (tier-2 wrap-DEKs variant) landing, a 5th \`null\` case ships, and the cast pattern compounds.

## Proposed fix

Tighten the type:

\`\`\`ts
interface UnlockedKeyring {
  // ...
  readonly kek: CryptoKey | null   // null for tier-3 / wrap-DEKs / unencrypted / session restore
  // ...
}
\`\`\`

Then audit every consumer of \`keyring.kek\` and add explicit null-checks:
- \`packages/hub/src/vault.ts:1450\` (cross-vault rewrap)
- \`packages/hub/src/team/keyring.ts:919\` (DEK propagation in grant)
- Any \`@noy-db/on-*\` package's verifier that calls \`subtle.wrapKey('raw', keyring.kek, ...)\`

## Why not in #26

#26 (tier-2 wrap-DEKs variant) preserves the existing cast pattern to keep the PR scope tight. This issue tracks the principled cleanup as a separate change.

## Pointer

The 4 existing cast sites work in practice — every code path that produces a \`null\` kek is one where downstream consumers don't reach for \`keyring.kek\`. The type tightening is mostly a hygienic exercise that prevents the next contributor from assuming \`kek\` is always present.

vLannaAi:
**In flight in PR #52** — release-prep branch for v0.1.0-pre.8.

The runtime contract has always allowed null for `keyring.kek`. Five call sites passed `null as unknown as CryptoKey`:

- `packages/hub/src/noydb.ts:123` — `createPlaintextKeyring` (unencrypted mode)
- `packages/hub/src/session/session.ts:251` — session restore
- `packages/hub/src/session/dev-unlock.ts:266` — dev unlock fixture
[...]

Issue #34, opened 2026-05-07
feat(hub): atomic db.recoverUser(userId, tempPhrase) primitive


vLannaAi:
## Problem

Peer-recovery (whether composed by consumers or eventually built into \`@noy-db/on-magic-link\` per #32) does:

```
1. db.revoke({ userId })   // delete the keyring file for the forgotten-phrase user
2. db.grant({ userId, passphrase: tempPhrase })   // re-create it with a fresh wrapping
```

These are **two writes**. If the issuer cancels, the network drops, the browser closes between them — the target user has no keyring file at all. They're locked out, and the issuing owner must re-issue.

Niwat ships this composed flow in \`app/auth/invite.ts:issuePeerRecovery\` and notes the partial-failure window in the docstring:

> Caveat: \`db.revoke\` followed by \`db.grant\` is two writes. If the user cancels between them, Owner 1's slot is missing entirely — they must re-issue. File upstream: \`db.recoverUser(userId, tempPhrase)\` atomic.

## Proposed primitive

```ts
await db.recoverUser({
  userId,
  displayName,    // optional — reuse existing if absent
  role,           // optional — reuse existing if absent
  passphrase: tempPhrase,
  // Optional factor proofs (gates this — see #33)
  factors: [...],
})
```

Hub does revoke + grant in a single atomic store transaction. On partial failure, the original keyring stays intact. On success, the new keyring's wrapping is under PBKDF2(tempPhrase) and the recipient can immediately call \`db.rotatePassphrase\` to choose their own phrase.

## Why medium priority

Doesn't block functionality — Niwat's compose-from-primitives version works most of the time. But "you must re-issue if the user closes the tab between two HTTP roundtrips" is the kind of edge case that bites once a year and erodes trust.

Pairs with #33 (owner-to-owner) and #32 (\`@noy-db/on-magic-link\`).

Tracked downstream at: niwat#34 phase 7 (\`issuePeerRecovery\` non-atomicity).

vLannaAi:
**Shipped in v0.1.0-pre.8** via PR #46.

`db.recoverUser(vault, options, factors?)` is now an atomic primitive that closes the partial-failure window:

\`\`\`ts
await db.recoverUser('acme', {
  userId: 'bob',
  passphrase: tempPhrase,
  // role + displayName optional; default to existing target's values
})
\`\`\`

[...]

Issue #3, opened 2026-05-01
by-tabs: duplicate RPC responses when 3+ tabs all run servePeerStore


vLannaAi:
## Summary

When wiring `@noy-db/by-tabs` with the recommended `peerStore` + `servePeerStore` pattern, every tab both serves and consumes the shared `BroadcastChannel`. With 2 tabs this works correctly — `BroadcastChannel` does not echo to the sender, so there is no self-loop. With 3+ tabs open simultaneously, each non-sending tab that runs `servePeerStore` will respond to an incoming RPC request independently, producing **duplicate responses** for the same request ID.

## Reproduction

```ts
// noydb.client.ts (Nuxt plugin — runs identically in every tab)
import { tabsChannel, isTabsChannelAvailable } from '@noy-db/by-tabs'
import { peerStore, servePeerStore } from '@noy-db/by-peer'
import { browserIdbStore } from '@noy-db/to-browser-idb'

const localStore = browserIdbStore({ prefix: 'niwat' })

if (isTabsChannelAvailable()) {
  const channel = tabsChannel({ name: 'niwat:vault' })
  servePeerStore({ channel, store: localStore })   // tab answers RPC requests
}

const db = await createNoydb({
  store: routeStore({
    default: localStore,
    sync: peerStore({ channel }),                  // tab issues RPC requests
  }),
  // ...
})
```

Open tab A, B, C. Tab A writes a record. Tab B and Tab C both call `servePeerStore` → both answer tab A's sync RPC with the same response. Depending on how `peerStore` handles duplicate responses for the same request ID, this may result in:

- Duplicate `put` calls on the receiving vault
- A race where the second response overwrites a locally-committed version
- Or silent discard — whichever `peerStore` does first is unclear from the current docs/types

## Expected behaviour

Either:
1. **Leader election**: only one tab acts as the server at a time (e.g. via `navigator.locks` or a `BroadcastChannel`-based lock), or
2. **Idempotent handling**: `peerStore` deduplicates responses by request ID so duplicate answers from N servers are safe, or
3. **Documentation**: the current behaviour (what happens on duplicate response) is documented so consumers can decide whether to implement their own leader election.

## Context

Discovered while wiring `by-tabs` for the [Niwat](https://github.com/vLannaAi/niwat) accounting app. Two-tab case works correctly in manual testing. The 3+ tab case is untested at scale — this issue is a heads-up rather than a confirmed data-corruption report.

## Environment

[...]

PR #42, opened 2026-05-07
feat: tier-2 wrap-DEKs unification (#26 Path C, #38 Option A)

### LINKED ISSUE #26: feat(hub): db.enrollPassword(ceremony) parallel to db.enrollWebAuthn

### LINKED ISSUE #38: fix(on-recovery): @pre.7 produces wrappedKEK entries; hub expects wrappedDeks

vLannaAi:
## Summary

Niwat's review of the post-pre.7 auth surface surfaced a structural asymmetry: tier-0 (paper recovery, `mintPaperRecoveryEntry`) and tier-3 (`@noy-db/on-pin`) both wrap the **DEK SET** under a per-credential AES-GCM key, while tier-2 (`@noy-db/on-password`) was the only tier wrapping the **KEK** directly. That asymmetry forced the tier-2 ceremony to require an extractable KEK — which the hub's `deriveKey` deliberately disallows — making `@noy-db/on-password` unreachable from a real consumer.

This PR unifies all three tiers around the **wrap-DEKs primitive**.

## Why this approach (Path C, not Path A)

An earlier round of design considered Path A — flip `extractable: true` on the KEK at `crypto.ts:72` (one-line change). It works, but leaves tier-2 as the architectural outlier. Path C eliminates the outlier entirely — every tier-0/2/3 ceremony now follows the same shape: serialize the DEK set, encrypt under a credential-derived key, store the ciphertext in a keyring slot or `_meta/recovery-paper` document.

Trade-off: tier-2 unlocks now produce `UnlockedKeyring` with `kek: null`. Sensitive ops (`enrollAuthenticator`, `rotatePassphrase`) require a tier-1 unlock anyway, matching how `@noy-db/on-pin` already works at tier 3. Niwat accepted this trade-off in the #26 comment.

### `@noy-db/hub` — minor
- `KeyringAuthenticator` is now a **discriminated union** over wrap-KEK and wrap-DEKs variants. Backward-compat: pre-pre.8 slots without `wrapKind` are treated as wrap-KEK at unlock.
- `EnrollAuthenticatorOptions` becomes the parallel union; `enrollAuthenticator` accepts either and persists the variant verbatim.
- `db.enrollRecovery` docstring updated — on-recovery now works end-to-end as the canonical example.

### `@noy-db/on-password` — minor (BREAKING)
- `enrollPasswordAuthenticator` wraps the DEK set under a PBKDF2-derived AES-GCM key (parallels `mintPaperRecoveryEntry`). No `keyring.kek` requirement.
- `verifyPasswordSlot` returns an `UnlockedKeyring` with `kek: null`. Drops the `materialize` callback — replaced by a single `keyring` reference for identity fields.
- `unwrapKekWithPassword` removed; replaced by `unwrapDeksWithPassword`.
- Pre-pre.8 wrap-KEK password slots throw `PasswordInvalidError` with a clear "re-enrol" message.

### `@noy-db/on-recovery` — minor (BREAKING)
- `generateRecoveryCodeSet({ deks, count })` (was: `{ kek, count }`). Delegates to hub's `mintPaperRecoveryEntry`.
- Removed: `unwrapKEKFromRecovery`, `wrapKEKForRecovery`, `deriveRecoveryWrappingKey`, `RecoveryCodeEntry` type.
- The package is now a thin code-generator + parser layer over the hub.

## Test plan

- [x] 2350 / 2350 tests pass (181 test files; 4 skipped IDB-perf benches)
- [x] Typecheck: 107 / 107 targets clean
- [x] Lint: 106 / 106 packages clean
- [x] New `pr1b-authenticator-shapes.test.ts` pins the discriminated union round-trip + coexistence in one keyring
- [x] Full on-password rewrite + legacy wrap-KEK rejection regression test
- [x] Full on-recovery rewrite + hub-delegated round-trip test
- [x] Showcase 26 updated to new API
- [ ] Niwat sanity-check: replace the vendored ~70 LOC + drop the synthetic `materialize` workaround for `verifyPasswordSlot`

## Branch base

This PR is based on `feat/pr1a-public-recovery-api` (PR #40). When PR #40 merges to main, retarget this PR to main — the diff against main should remain the architectural changes only.

## Follow-up

#41 tracks the `UnlockedKeyring.kek` type tightening (`CryptoKey | null`) — deliberately scoped out. The runtime already lies in 5 places (4 pre-existing + 1 new wrap-DEKs verifier); the audit is its own PR.

Closes #26
Closes #38

### DIFF
--- a/packages/hub/__tests__/pr1b-authenticator-shapes.test.ts
+++ b/packages/hub/__tests__/pr1b-authenticator-shapes.test.ts
@@ -1,0 +1,130 @@
+/**
+ * PR1b — `KeyringAuthenticator` discriminated union for wrap-KEK and
+ * wrap-DEKs slot variants (#26 Path C).
+ *
+ * Pins the on-disk + type-level contract:
+ *   - WebAuthn / OIDC slots use `wrapped_kek` (legacy / wrapKind absent
+ *     or 'kek'); the existing format keeps working unchanged.
+ *   - Password slots use `wrapped_deks` + `iv` + `wrapKind: 'deks'`;
+ *     they sidestep the non-extractable-KEK constraint by wrapping the
+ *     DEK set under an AES-GCM key, mirroring `mintPaperRecoveryEntry`
+ *     and `@noy-db/on-pin`'s tier-3 pattern.
+ *
+ * The test does not exercise the actual unlock flow — that lives in
+ * `@noy-db/on-password`'s tests after the package refactor. This test
+ * pins the hub-level dispatch + persistence shape only.
+ */
+import { describe, it, expect } from 'vitest'
+import type { NoydbStore, EncryptedEnvelope, KeyringAuthenticator } from '../src/types.js'
+import { createOwnerKeyring, loadKeyring, persistKeyring } from '../src/team/keyring.js'
+import { enrollAuthenticator } from '../src/team/authenticators.js'
+import { generateDEK } from '../src/crypto.js'
+
+function inlineMemory(): NoydbStore {
+  const store = new Map<string, Map<string, Map<string, EncryptedEnvelope>>>()
+  function gc(c: string, col: string) {
+    let comp = store.get(c)
+    if (!comp) { comp = new Map(); store.set(c, comp) }
+    let coll = comp.get(col)
+    if (!coll) { coll = new Map(); comp.set(col, coll) }
+    return coll
+  }
+  return {
+    name: 'inline-memory',
+    async get(c: string, col: string, id: string) { return gc(c, col).get(id) },
+    async put(c: string, col: string, id: string, env: EncryptedEnvelope) { gc(c, col).set(id, env) },
+    async delete(c: string, col: string, id: string) { gc(c, col).delete(id) },
+    async list(c: string, col: string) { return [...gc(c, col).keys()] },
+    async loadAll() { return {} },
+    async saveAll() {},
+    capabilities: { casAtomic: true, auth: { kind: 'none' } },
+  } as unknown as NoydbStore
+}
+
+const PHRASE = 'correct horse battery staple printer toaster'
+
+describe('KeyringAuthenticator wrap-KEK / wrap-DEKs (#26 Path C)', () => {
+  it('round-trips a wrap-KEK slot through _keyring (legacy WebAuthn shape)', async () => {
+    const store = inlineMemory()
+    const keyring = await createOwnerKeyring(store, 'acme', 'alice', PHRASE)
+    keyring.deks.set('invoices', await generateDEK())
+    await persistKeyring(store, 'acme', keyring)
+
+    const next = await enrollAuthenticator(store, 'acme', keyring, {
+      id: 'webauthn-01',
+      method: 'webauthn',
+      meta: { credentialId: 'cred-base64', prfUsed: true },
+      wrapped_kek: 'YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWE=',
+    })
+    expect(next.authenticators).toHaveLength(1)
+
+    const reloaded = await loadKeyring(store, 'acme', 'alice', PHRASE)
+    expect(reloaded.authenticators).toHaveLength(1)
+    const slot = reloaded.authenticators[0]!
+    // Discriminator absence (or 'kek') → wrap-KEK
+    expect(slot.wrapKind).toBeUndefined()
+    if (slot.wrapKind !== 'deks') {
+      expect(slot.wrapped_kek).toBe('YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWE=')
+      // wrap-DEKs fields must NOT be present on a wrap-KEK slot
+      expect((slot as Partial<KeyringAuthenticator & { wrapped_deks?: string; iv?: string }>).wrapped_deks).toBeUndefined()
+    } else {
+      throw new Error('expected wrap-KEK slot, got wrap-DEKs')
+    }
+  })
+
+  it('round-trips a wrap-DEKs slot through _keyring (new password shape)', async () => {
+    const store = inlineMemory()
+    const keyring = await createOwnerKeyring(store, 'acme', 'alice', PHRASE)
+    keyring.deks.set('invoices', await generateDEK())
+    await persistKeyring(store, 'acme', keyring)
+
+    const next = await enrollAuthenticator(store, 'acme', keyring, {
+      id: 'password-daily',
+      method: 'password',
+      wrapKind: 'deks',
+      wrapped_deks: 'Y2lwaGVydGV4dC1iYXNlNjQ=',
+      iv: 'aXYtYmFzZTY0LTEy',
+      meta: { salt: 'c2FsdC1iYXNlNjQ=', minLength: 12 },
+    })
+    expect(next.authenticators).toHaveLength(1)
+
+    const reloaded = await loadKeyring(store, 'acme', 'alice', PHRASE)
+    expect(reloaded.authenticators).toHaveLength(1)
+    const slot = reloaded.authenticators[0]!
+    expect(slot.wrapKind).toBe('deks')
+    if (slot.wrapKind === 'deks') {
+      expect(slot.wrapped_deks).toBe('Y2lwaGVydGV4dC1iYXNlNjQ=')
+      expect(slot.iv).toBe('aXYtYmFzZTY0LTEy')
+      // wrap-KEK field must NOT be present on a wrap-DEKs slot
+      expect((slot as Partial<KeyringAuthenticator & { wrapped_kek?: string }>).wrapped_kek).toBeUndefined()
+    } else {
+      throw new Error('expected wrap-DEKs slot, got wrap-KEK')
+    }
+  })
+
+  it('coexists — both shapes can live in the same keyring', async () => {
+    const store = inlineMemory()
+    const keyring = await createOwnerKeyring(store, 'acme', 'alice', PHRASE)
+    await persistKeyring(store, 'acme', keyring)
+
+    const afterWebAuthn = await enrollAuthenticator(store, 'acme', keyring, {
+      id: 'webauthn-01',
+      method: 'webauthn',
+      meta: { credentialId: 'cred-1' },
+      wrapped_kek: 'd2VicGF5bG9hZA==',
+    })
+    const afterPassword = await enrollAuthenticator(store, 'acme', afterWebAuthn, {
+      id: 'password-daily',
+      method: 'password',
+      wrapKind: 'deks',
+      wrapped_deks: 'cHdkLXdyYXBwZWQ=',
+      iv: 'cHdkLWl2',
+      meta: { salt: 'cHdkLXNhbHQ=', minLength: 12 },
+    })
+    expect(afterPassword.authenticators).toHaveLength(2)
+
+    const reloaded = await loadKeyring(store, 'acme', 'alice', PHRASE)
+    const kinds = reloaded.authenticators.map((a) => a.wrapKind ?? 'kek')
+    expect(kinds).toEqual(['kek', 'deks'])
+  })
+})
--- a/packages/hub/src/noydb.ts
+++ b/packages/hub/src/noydb.ts
@@ -1424,10 +1424,15 @@ export class Noydb {
    * showCodesToUser(codes)
    * ```
    *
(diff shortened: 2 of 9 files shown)


vLannaAi:
**Niwat consumer-side review** — architecture LGTM, six concrete points worth addressing in this PR or a fast follow-up.

This is exactly Path C as proposed in #26. Unifying tier-0 / tier-2 / tier-3 around the wrap-DEKs primitive eliminates the architectural outlier and the extractable-KEK constraint in one move. on-recovery becomes a thin code-generator + parser layer over hub's `mintPaperRecoveryEntry` — zero crypto duplication. `pr1b-authenticator-shapes.test.ts` covers wrap-KEK round-trip (legacy), wrap-DEKs round-trip (new), and coexistence in one keyring. Backward-compat handling at unlock (`wrapKind?: 'kek'` defaulting to wrap-KEK for legacy slots) is documented at the type level and called out in the docstring.

### Non-blocking review comments

**1. Type-level XOR is overstated.** PR body and `KeyringAuthenticator` docstring both claim:

> The type-level XOR enforces "exactly one of `wrapped_kek` / `wrapped_deks` is present" — a structural guarantee that the runtime dispatch is safe.

The union as written doesn't structurally exclude crossover. This compiles today:

```ts
const slot: KeyringAuthenticator = {
  ...base,
  wrapKind: 'kek',
  wrapped_kek: 'aaa',
  wrapped_deks: 'bbb',  // not rejected
  iv: 'ccc',
}
```

True XOR needs `wrapped_deks?: never; iv?: never` on `KeyringAuthenticatorWrappingKEK` (and `wrapped_kek?: never` on `KeyringAuthenticatorWrappingDEKs`). Runtime dispatch is fine because every consumer narrows on `wrapKind` first, so this is correctness rather than safety — but the docstring should soften the claim or the types should add the guards.

**2. `kek: null as unknown as CryptoKey` cast in `verifyPasswordSlot`** — known per the PR body ("the runtime already lies in 5 places... the audit is its own PR (#41)"). Suggest a `// TODO(#41): drop the cast once UnlockedKeyring.kek is CryptoKey | null` inline so future readers don't think it's accidental.

**3. `verifyPasswordSlot`'s new `keyring` parameter has cold-start ergonomics issues.**

```ts
verifyPasswordSlot(slot, password, { keyring: UnlockedKeyring })
```

Works fine for **post-lock re-unlock** (db is alive, consumer calls `await db.getKeyring(vault)`) and for **enrol-while-unlocked** (settings page). But for **cold-start tier-2 unlock** via `createNoydb({ getKeyring })`, the consumer doesn't have a keyring yet — they're trying to GET one. They'd have to `loadKeyringFile(store, vault, userId)` directly, build a partial `UnlockedKeyring`-shaped object with `{ deks: empty Map, kek: null }`, and pass that. Awkward shape contract.

Two cleaner alternatives worth considering before merge:

- **Accept identity fields explicitly** — `{ userId, displayName, role, permissions, authenticators, salt }` instead of `keyring`. Decouples from a fully-formed `UnlockedKeyring`.
- **Have the verifier load identity itself** — pass `{ store, vault, userId }` and let `verifyPasswordSlot` read the keyring file directly (it already needs the salt etc. via `slot.meta`).

At minimum a docstring on how cold-start should construct the reference. Niwat's planned tier-2b integration would be the proving ground for the cold-start path.

**4. Legacy-rejection test asserts class but not message.**

```ts
await expect(unwrapDeksWithPassword(legacy, 'pwd'))
  .rejects.toBeInstanceOf(PasswordInvalidError)
```

The recovery instruction is the user-facing message ("Pre-pre.8 wrap-KEK password slots are no longer supported — re-enrol via enrollPasswordAuthenticator"). A future refactor that throws `PasswordInvalidError` for a different reason would silently regress that UX. Suggest:

```ts
.rejects.toMatchObject({ message: expect.stringMatching(/wrap-DEKs|re-enrol/) })
```

**5. on-recovery scope just shrunk — package consolidation question.** After this PR, on-recovery is ~110 LOC of Base32 + checksum + delegation. Worth considering whether it should move under `@noy-db/hub/recovery-codes` (subpath, like `@noy-db/hub/blobs`). Not a blocker for this PR; flag for follow-up since the package's scope changed dramatically.

**6. `_meta/recovery-paper` ↔ `KeyringAuthenticator` wrap-DEKs duplication.** Two distinct on-disk formats now carry the same wrap-DEKs primitive (recovery entries at `_meta/recovery-paper`, password slots in `KeyringFile.authenticators`). Consider exporting a shared `WrappedDeksBlob = { wrapped_deks: string; iv: string; salt: string }` type so future on-* packages and the hub itself reference one canonical shape.

### Niwat consumer cleanup, when this lands

- Drop the vendored ~70 LOC in `niwat-app/app/auth/recovery.ts`; import `mintPaperRecoveryEntry` from `@noy-db/hub` directly (already prepared for this — the vendored block is fenced with a comment pointing at #39).
- Tier-2b daily password is now unblocked (was deferred in #34 phase 2). Will exercise the cold-start `verifyPasswordSlot` path and surface point #3 above as a real consumer concern.

vLannaAi:
Thanks for the thorough review. Addressing all six:

### In this PR (commit `8852974`)

**Point 1 — type-level XOR overstated.** Fixed. Added `wrapped_deks?: never; iv?: never` on `KeyringAuthenticatorWrappingKEK` and `wrapped_kek?: never` on `KeyringAuthenticatorWrappingDEKs`. Crossover slots now fail to compile. Existing call sites (which never set the wrong fields) compile unchanged.

**Point 2 — `// TODO(#41)` comment.** Added next to the cast site in `verifyPasswordSlot`.

**Point 3 — `verifyPasswordSlot` cold-start ergonomics.** Took the cleaner path you flagged. Replaced `{ keyring }` with `{ store, vault, userId }`:

```ts
verifyPasswordSlot(slot, password, { store, vault: 'acme', userId: 'alice' })
```

The verifier now reads identity directly from `_keyring/<userId>` — same pattern `team/rotate-recover.ts:191-195` already uses for paper recovery (the keyring envelope's plaintext header is intentionally readable without a key, since the sync engine + grant flow need it). One disk read per unlock; works in both cold-start and warm contexts; one uniform API.

Added a missing-keyring rejection test (`PasswordInvalidError` with a `verifyPasswordSlot: no keyring found at "vault/_keyring/userId"` message). Tier-2b's cold-start should now drop in cleanly.

**Point 4 — legacy-rejection test asserts message.** Strengthened from `.rejects.toBeInstanceOf(PasswordInvalidError)` to `.rejects.toMatchObject({ name, message: /wrap-DEKs|re-enrol/ })`. A future refactor that throws `PasswordInvalidError` for a different reason would now fail the test rather than silently regressing the user-facing "re-enrol" instruction.

### Filed as follow-ups

**Point 5 — on-recovery package consolidation** → #43. Decision criteria written in: if on-recovery stays at ~110 LOC indefinitely, fold under `@noy-db/hub/recovery-codes`; if it grows (audit helpers, rate limiter, printable templates), keep as its own package.

**Point 6 — `WrappedDeksBlob` shared type** → #44. Net-neutral until a third on-* package adopts wrap-DEKs. Capturing the unification opportunity now while the PR #42 design is fresh.

### Verification (post-rebase)

- 2351 / 2355 tests pass (+1 missing-keyring test; 4 skipped IDB-perf benches)
- Typecheck: 107 / 107 clean
- 18 affected tests green: 9 on-password, 6 enroll-webauthn (legacy wrap-KEK still compiles + unlocks under the new XOR guards), 3 hub shape round-trip

PR #42 retargeted to `main` after PR #40 merged; the rebase dropped the duplicate PR1a commit so the diff against main now shows only the architectural changes. CI re-running.

Issue #147, opened 2026-05-20
feat(derivations): give derive() the same ReadOnlyVaultFacade that guards have


vLannaAi:
## Summary

Give `withDerivation`'s `derive(source)` function the same `ReadOnlyVaultFacade` that `withGuard`'s `check(incoming, ctx)` already has. Today, `derive` only sees the source record — no sibling reads. Any field the derivation needs from another collection has to be **denormalised onto the source row at write time and frozen by a paired `withGuard.frozenFields`** to prevent drift. That works, but it's a pair of primitives doing what one primitive plus `ctx.vault` could do alone.

## Proposed API

Mirror the guard surface exactly:

```ts
// packages/hub/src/derivations/types.ts
export interface DerivationContext {
  vault: ReadOnlyVaultFacade   // same one guards already get
}

derive: (source: TSource, ctx: DerivationContext) => Promise<TOutputs> | TOutputs
```

`DerivationContext` starts narrow (vault only). No transaction handle, no write capability — same restrictions as `GuardContext`. Strategy hash incorporates whatever sibling reads happen, so determinism semantics don't change (consumer is responsible for the function being deterministic given the same inputs).

## Worked example from niwat-app

`paymentAllocation → receipt` derivation needs three things from sibling collections:

| Field | Lives on | Why derive needs it |
|------|----------|---------------------|
| `payment.paymentDate` | `payments` | Receipt's `issuedAt` (RCT-DATE-001) — drives which VAT month the receipt belongs to |
| `bill.servicesNetUnpaid`, `bill.expenseUnpaid` | `invoices` | Inputs to `splitPaymentForReceipt` — slices `appliedAmount` into services/VAT/expense buckets (RCT-ALLOC-001) |
| `bill.clientId` | `invoices` | Required field on Receipt for query indexing |

**Today's workaround** (no `ctx.vault` in derive):

```ts
interface AllocationWithEmbedded extends PaymentAllocation {
  // Embedded at write time by consumer (in payments store)
  _paymentDate: string                    // = payment.paymentDate
  _bill: {
    id: string
    clientId: string
    servicesNetRemaining: number          // = bill.servicesNetUnpaid AT ALLOCATION TIME
    expenseRemaining: number              // = bill.expenseUnpaid
  }
}

// Paired withGuard so the embedded fields can't drift
withGuard({
  collection: 'paymentAllocations',
  frozenFields: { when: () => true, fields: ['_paymentDate', '_bill'] },
})
```

[...]

Issue #146, opened 2026-05-20
feat(guards): expose .query() on ReadOnlyVaultFacade for aggregating checks


vLannaAi:
## Summary

`withGuard.check` runs at put-time with `ctx.vault: ReadOnlyVaultFacade` — but the facade only exposes `.get(id)` and `.list()`. Any cross-record invariant that needs `Σ`/`max`/`groupBy` over sibling rows has to `.list()` the whole collection in userland and reduce manually. This is fine semantically (the data is all there) but it's:

1. **Inefficient** — `.list()` decrypts every record in the collection on every put.
2. **Repetitive** — every consumer writes the same `(rows) => rows.filter(...).reduce(...)` boilerplate.
3. **A duplicate of the query DSL** — `sum`, `count`, `groupBy`, `Aggregation` already ship in `packages/hub/src/aggregate/`, but they're not reachable from inside a guard.

## Proposed API

Add `.query()` to `ReadOnlyVaultFacade`, returning the same query builder the rest of the library uses:

```ts
// packages/hub/src/guards/types.ts
export interface ReadOnlyVaultFacade {
  collection<T = unknown>(name: string): {
    get(id: string): Promise<T | null>
    list(): Promise<T[]>
    query(): QueryBuilder<T>   // NEW — same QueryBuilder as the top-level vault
  }
}
```

Guards can then do:

```ts
check: async (incoming, ctx) => {
  const sum = await ctx.vault
    .collection<PaymentAllocation>('paymentAllocations')
    .query()
    .where('paymentId', incoming.paymentId)
    .aggregate({ total: sum('appliedAmount') })
  // ...
}
```

Read-only: `query()` only exposes `.where()` / `.aggregate()` / `.groupBy()` / `.join()` paths that don't write. (The existing `QueryBuilder` is already read-only — there's no `.update()` or `.delete()` on it — so this is more about typing than restriction.)

## Worked example from niwat-app

`PAYALLOC-SUM-001` — sum of allocation amounts must not exceed the payment:

> "For every Payment `P`, `Σ allocation.appliedAmount` over allocations with `paymentId === P.id` must be ≤ `P.amount`."

**Today** (without query in facade):

```ts
withGuard({
  collection: 'paymentAllocations',
  check: async (incoming, ctx) => {
    // Decrypt the entire paymentAllocations collection
    const all = await ctx.vault.collection<PaymentAllocation>('paymentAllocations').list()
[...]

Issue #145, opened 2026-05-20
feat(guards): withGuard.onDelete — reject delete based on record state


vLannaAi:
## Summary

`withGuard.check` only runs on `Collection.put`. There's no equivalent hook on `Collection.delete`. Today, "reject delete based on record state" can only be expressed via the `refEnforcer` (FK cascade/strict/warn) — which is the wrong shape for rules that depend on the record's *own* fields rather than inbound references.

## Proposed API

Extend `GuardStrategy` with an optional `onDelete` predicate:

```ts
export interface GuardStrategy<T extends Record<string, unknown>> {
  collection: string
  check?: (incoming: T, ctx: GuardContext<T>) => Promise<void> | void
  onDelete?: (existing: T, ctx: GuardContext<T>) => Promise<void> | void  // NEW
  frozenFields?: { /* ... */ }
  amendment?: { /* ... */ }
}
```

Semantics:
- `onDelete(existing, ctx)` runs inside `Collection.delete` **before** the adapter delete and the ledger append (mirroring where `check` runs on put).
- Throwing from `onDelete` cancels the delete (no partial state, no ledger entry).
- `ctx` is the same `GuardContext` (vault facade, userId, role).
- Like `check`, `onDelete` is bypassed inside `db.transaction({ amendment: true })` for admin/owner — amendments are the unlock primitive.

## Worked examples from niwat-app

Two niwat rules need this primitive:

### `BILL-DELETE-001` — only unsent / cancelled bills may be deleted

> "Bills can only be deleted while in `new`, `draft`, or `cancelled` status. Sent / paid bills must stay in the record (paid bills have receipts attached)."

Today the consumer's `deleteInvoice` checks `status` imperatively. The rule wants this structurally enforced so any future code path that calls `db.collection('invoices').delete(id)` gets the same check. Without `onDelete`:

```ts
// Consumer has to wrap every delete call manually
async function deleteInvoice(id) {
  const inv = await db.invoices.get(id)
  if (!['new', 'draft', 'cancelled'].includes(inv.status)) {
    throw new Error('Cannot delete invoice in status ' + inv.status)
  }
  await db.invoices.delete(id)
}
// Direct calls to `db.invoices.delete(id)` from elsewhere bypass the check.
```

With `onDelete`:

```ts
withGuard({
  collection: 'invoices',
  onDelete: (existing) => {
    if (!['new', 'draft', 'cancelled'].includes(existing.status)) {
[...]

Issue #130, opened 2026-05-18
fix(hub): investigate ~30-42% gzipped bundle regression introduced at guards T6


vLannaAi:
## Context

The guards feature batch (#123) introduced an unexpected bundle-size regression:

| Scenario  | Baseline gz | After guards gz | Delta |
|-----------|-------------|------------------|-------|
| floor     | 30,546      | 43,435           | +42.2% |
| history   | 33,987      | 47,408           | +39.5% |
| analytics | 32,860      | 45,925           | +39.8% |
| all-on    | 47,949      | 62,125           | +29.6% |

Each scenario gained ~12-14 KB gzipped. The guards source itself (errors + types + facade + factory + registry + executor + barrel) is ~380 lines of TypeScript — expected to add 2-4 KB gzipped, not 12+ KB.

## Bisect

Independent verification on the `feat/guards` branch traced the regression to **commit `131c381` (T6 — `feat(hub): export withGuard + guard types + error classes`)**. Earlier guards commits (T1-T5: error classes, types, facade, factory, registry, executor) did not move the bundle. T6b (`@noy-db/hub/guards` subpath registration) added only +13 bytes gzipped.

## Hypothesis

The `packages/hub/src/index.ts` re-exports from `./guards/index.js` — and the guards barrel re-exports `GuardRegistry`, `GuardExecutor`, `ReadOnlyVaultFacade` in addition to `withGuard`. Even though the main index only re-exports `withGuard`, tsup may be including the entire guards barrel transitively when bundling `index.ts`. `splitting: true` should tree-shake the unused symbols, but apparently doesn't here.

## Investigation steps

1. Inspect `dist/index.js` after build — is `GuardRegistry` / `GuardExecutor` source actually present?
2. Compare with how `tx` does it — `runTransaction` is also re-exported from the main barrel, does it also pull in `TxContext` / `TxCollection` etc.?
3. If tree-shaking fails: change the main `index.ts` re-export to import directly from `./guards/with-guard.js` instead of via the barrel
4. Re-run `pnpm --filter=@noy-db/hub bundle-check` after each attempt

## Acceptance

- Bundle floor returns to within ~5% of pre-guards baseline (≤ 32 KB gz)
- `@noy-db/hub/guards` subpath import still resolves
- All guards tests still pass

## References

- Branch `feat/guards` (commit `bba866a` HEAD)
- Surfaced by code-review of T6 and T6b in the pre.11 milestone execution

PR #148, opened 2026-05-20
feat(hub): Guards/Derivations v1.5 — #144 #145 #146 #147

### LINKED ISSUE #144: feat(derivations): optional outputs in withDerivation (return null = skip emission)
## Summary

`withDerivation` v1 requires every declared output to be returned on every `derive` call. There's no way to express "this source row sometimes produces an output and sometimes doesn't." For consumers where the source has a discriminator field that determines whether an output should exist, this forces either a sentinel-record pattern (write a "skip" record and filter at read) or splitting the source into two collections artificially.

## Proposed API

Allow each output declaration to be optional, and let `derive` return `null` for that key:

```ts
// packages/hub/src/derivations/types.ts
export interface OutputSpec {
  shape: 'record'
  collection: string
  optional?: boolean  // NEW — when true, derive may return null for this key
}

derive: (source) => Promise<{
  [K in keyof TOutputs]: TOutputs[K] | (Spec[K]['optional'] extends true ? null : never)
}>
```

Semantics:
- `null` return for an optional output → no write happens; if a previous derivation produced an output, it's **deleted** (mirrors the empty-aggregate-group question on #142).
- Required outputs (default) must still return a record; returning `null` throws `DerivationOutputShapeError`, same as today.

## Worked example from niwat-app

`paymentAllocation → receipt`. Most allocations produce a tax receipt, but **expense-only allocations don't** — RCT-TRIGGER-001 says receipts only exist for services-touching allocations (no VAT-bearing revenue = no receipt to issue).

Today's allocation flow has to branch:
- Allocation covers services portion of a bill (`servicesNetPortion > 0`) → emit Receipt
- Allocation covers only expense reimbursement (`servicesNetPortion === 0`) → no Receipt

Three workarounds on current v1, all bad:

1. **Sentinel "skip" receipt.** Write a Receipt record with `paidServicesNet: 0`, `paidServicesVat: 0`, and a `skipped: true` flag. Filter at read. **Cost:** pollutes the receipts collection; tax-audit listings (RD inspector) would see ghost rows; monthly receipt sequence numbers (RCT-NUM-001) get burned on skipped rows.

2. **Two source collections.** Split `paymentAllocations` into `serviceAllocations` (derives Receipt) and `expenseAllocations` (no derivation). **Cost:** the bridge-table identity is one thing — splitting it for an enforcement quirk forces every consumer of "allocations for payment X" to read two collections and union them. Cascades into views.

3. **Imperative path stays alongside the derivation.** Keep `writeAllocationWithReceipt` in the consumer, only call it for services-touching allocations. **Cost:** the rule was supposed to be structurally enforced; instead the structural enforcement only applies *if* the consumer routes through it. The "did the consumer remember?" risk is exactly what `withDerivation` was supposed to remove.

With `optional: true`:

```ts
withDerivation<PaymentAllocation, { receipt: Receipt }>({
  source: 'paymentAllocations',
  outputs: {
    receipt: { shape: 'record', collection: 'receipts', optional: true },
  },
  derive: (alloc) => {
    if (alloc.servicesNetPortion <= 0) {
      return { receipt: null }   // expense-only — no receipt
    }
    return { receipt: { /* ... */ } }
  },
  lifecycle: 'eager',
  strict: true,
})
```

Single declarative block; no sentinel records; no source-collection split; the discriminator (`servicesNetPortion > 0`) lives where it belongs (inside derive, not in a consumer wrapper).

## Edge: what about transition from "had output" to "should not"?

If a source row produced an output, then is updated such that the output should no longer exist (e.g., allocation re-targeted to an expense-only bill), the executor should **delete** the existing output row in the same transaction as the source update.

This is the same semantic question MVs face on empty groups (#142). Suggest resolving both consistently: "transitioning to no-output = delete" by default, with an opt-out for consumers who want stale outputs to persist.

## Scope estimate

Small (~50 LoC):
- Add `optional?: boolean` to `OutputSpec`
- Update executor to allow `null` return for optional outputs without throwing `DerivationOutputShapeError`
- Update registry's persist step to delete existing output when derive returns `null` for an optional key
- Update strategy-hash to include the `optional` flag (optional bit affects determinism contract)

vLannaAi:
## Summary

Four fast-follow refinements to the Guards + Derivations subsystems that shipped in pre.11. All emerged from first-consumer use; they're the v1.5 strand of milestone pre.14, landing before the larger Dim 14 v2 (MV) work so MV inherits a hardened \`ReadOnlyVaultFacade\` foundation.

| Commit | Issue | What |
|---|---|---|
| 02557a2 | #146 | \`.query()\` on \`ReadOnlyVaultFacade\` — guards can aggregate sibling rows via the same chainable builder used elsewhere, no more \`.list()\` + manual reduce |
| 8249a1a | #147 | \`derive(source, ctx)\` — derivations get the same \`ReadOnlyVaultFacade\` guards have; sibling reads without denormalisation + frozen-fields pairing |
| df405d1 | #145 | \`withGuard.onDelete\` — dedicated delete-time hook; \`check\` becomes put-only |
| 973b179 | #144 | \`withDerivation\` optional outputs — \`null\` = skip emission + tombstone prior |

## Order rationale

#146 → #147 → #145 → #144 lands #146 + #147 (the shared-facade work) together so the abstraction settles in one review pass, then the independent #145 and #144 stack on top.

## Breaking surface changes (pre-1.0)

- **#145** — \`check\` no longer fires on \`Collection.delete\`. Any guard that previously gated deletes via \`check\` + \`ctx.existing\` must migrate to \`onDelete(existing, ctx)\`. The semantic split is explicit: \`check\` = "validate a write," \`onDelete\` = "validate a removal."
- **#147** — \`derive: (source) => ...\` widened to \`derive: (source, ctx) => ...\`. TypeScript's function bivariance keeps narrower existing signatures compatible at compile time, so most consumers see no diff unless they explicitly use \`ctx\`.

## Verification

- 1508 hub tests pass (was 1496 pre-PR · +12 new across the four issues + one regression guard)
- typecheck clean · lint clean
- Bundle within tolerance: floor +3.5% gz, all-on +1.5% gz (\`bundle-check\` ✓ on all four scenarios). The pre.12 floor-bundle hardening (#130) survives — \`Query\` and \`ReadOnlyVaultFacade\` are type-only or dynamic-imported.

## Subsystem docs

- \`docs/subsystems/guards.md\` — four-primitive table, \`onDelete\` semantics, \`.query()\` surface on facade
- \`docs/subsystems/derivations.md\` — \`derive(source, ctx)\` shape, optional outputs section with tombstone semantics
- \`ROADMAP.md\` — pre.14 marked in-flight as bundled MV + v1.5 milestone

## Test plan

- [x] \`pnpm --filter @noy-db/hub typecheck\`
- [x] \`pnpm --filter @noy-db/hub lint\`
- [x] \`pnpm --filter @noy-db/hub test\` — 1508 / 1508
- [x] \`pnpm --filter @noy-db/hub bundle-check\` — all scenarios green
- [ ] Spot-check the breaking surface changes (#145, #147) against any consumer in flight before merge

Closes #144 #145 #146 #147.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -9,6 +9,24 @@
 - **Showcase + recipe coverage.** A runnable end-to-end test for every `with*()` strategy and every storage destination so adopters pick a backend by reading working code.
 - **`by-*` session-share family.** `@noy-db/by-peer` (WebRTC, renamed from `@noy-db/p2p`) and `@noy-db/by-tabs` (BroadcastChannel multi-tab sync) shipped together with the family debut. Next: `@noy-db/by-server` (WebSocket / SSE relay) and `@noy-db/by-room` (Liveblocks / Yjs y-websocket).
 
+## In flight
+
+**`0.1.0-pre.14` — Dim 14 v2 (MV) + Guards/Derivations v1.5.** Bundled milestone for two related strands:
+
+- **Guards/Derivations v1.5** — fast-follow refinements to the pre.11 primitives driven by first-consumer use. Landing first to give MV a hardened `ReadOnlyVaultFacade` foundation:
+  - [#146](https://github.com/vLannaAi/noy-db/issues/146) `withGuard`: expose `.query()` on `ReadOnlyVaultFacade` for aggregating checks.
+  - [#147](https://github.com/vLannaAi/noy-db/issues/147) `withDerivation`: give `derive(source, ctx)` the same `ReadOnlyVaultFacade` guards have.
+  - [#145](https://github.com/vLannaAi/noy-db/issues/145) `withGuard.onDelete` — reject delete based on record state.
+  - [#144](https://github.com/vLannaAi/noy-db/issues/144) `withDerivation` optional outputs (`null` = skip emission).
+- **Dim 14 v2 — `withMaterializedView`** — collection-level query derivation (query → materialized collection). Extends `withDerivation` v1 (#129):
+  - [#142](https://github.com/vLannaAi/noy-db/issues/142) v2 design spec (blocking).
+  - [#143](https://github.com/vLannaAi/noy-db/issues/143) implementation epic.
+  - Deferred to v3: streaming MVs (Dim 12), MapReduce views, scheduled refresh (Dim 11 hooks), rendered/server-authoritative views.
+
+Execution order: v1.5 (#146 → #147 → #145 → #144) → spec (#142) → MV epic (#143).
+
+**`0.1.0-pre.13` — Real-provider showcase batch (Apple/Google/LINE).** Deferred (env-gated, parked 2026-05-20). Showcase coverage is a 1.0 gate, not a pre-release gate; the six issues stay open at `priority: low` to close opportunistically when real-provider creds are in hand: [#64](https://github.com/vLannaAi/noy-db/issues/64), [#65](https://github.com/vLannaAi/noy-db/issues/65), [#73](https://github.com/vLannaAi/noy-db/issues/73), [#74](https://github.com/vLannaAi/noy-db/issues/74), [#75](https://github.com/vLannaAi/noy-db/issues/75), [#76](https://github.com/vLannaAi/noy-db/issues/76).
+
 ## Recently shipped (and what's deferred)
 
 The following capabilities are now in main:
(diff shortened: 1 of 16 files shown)


vLannaAi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

vLannaAi:
### Code review

Found 1 issue (re-reviewed against niwat's enforcement plan; deeper lens after the prior pass):

1. Derivation tombstone for `optional` outputs calls public `Collection.delete`, which now fires user `onDelete` guards on the output collection — blocks the exact combination of #144 + #145 niwat plans to ship (a `paymentAllocation → receipt` derivation with `optional: true`, plus `onDelete: () => { throw ReceiptDeletionForbiddenError }` on `receipts` for `RCT-CANCEL-001`). System-internal tombstones need a bypass path; no test covers the combination.

https://github.com/vLannaAi/noy-db/blob/b2c4e9bd87ed6aa252ad68b899a67cd620329e58/packages/hub/src/collection.ts#L1407-L1428

(Lazy-path equivalent at https://github.com/vLannaAi/noy-db/blob/b2c4e9bd87ed6aa252ad68b899a67cd620329e58/packages/hub/src/derivations/stale.ts#L134-L145.)

Follow-ups posted on related issues:
- #144 comment (https://github.com/vLannaAi/noy-db/issues/144#issuecomment-4497517367) — full niwat use case + lazy-tombstone `txCtx._executed` registration asymmetry
- #145 comment (https://github.com/vLannaAi/noy-db/issues/145#issuecomment-4497517377) — `RCT-CANCEL-001` "unconditional throw" pattern silently has an amendment-bypass gap; spec question on pairing `onDelete` with `amendment.invariant`
- #142 comment (https://github.com/vLannaAi/noy-db/issues/142#issuecomment-4497517362) — same "system-internal delete vs user onDelete" question applies to MV refresh; spec should resolve consistently across derivations + MV

vLannaAi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

vLannaAi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

Issue #143, opened 2026-05-19
feat(hub): withMaterializedView — Dim 14 v2 implementation (epic)


vLannaAi:
## Overview

Implementation epic for the v2 MV strategy + the two primitives pulled forward during niwat-review (`declaredDeterministicPredicates`, `withOverlayedView` read-shadow variant).

**Spec landed:** [`docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md`](https://github.com/vLannaAi/noy-db/blob/main/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md) (#142 closed via PR #149).

## Sub-issues

Six PR-sized chunks covering the spec's 16-step sequencing:

- [ ] #150 — MV foundation (steps 1-6): factory, dependency analyzer, registry, executor, `_materializedFrom` envelope, `Collection.put` hook
- [ ] #151 — MV lazy lifecycle + `vault.refreshView` (steps 7-8)
- [ ] #152 — MV correctness: partition, `onEmpty`, ceiling, strict-mode rollback (steps 9-12)
- [ ] #153 — `declaredDeterministicPredicates` + `.wherePredicate(name, ctx?)` (step 13)
- [ ] #154 — `withOverlayedView` read-shadow primitive (steps 14-15)
- [ ] #155 — Showcases + recipes + subsystem doc + `features.yaml` (step 16)

## Sequencing

Steps 1-12 (sub-issues #150-#152) form the MV-only critical path and should land first. #153 + #154 are independent of each other and depend only on the foundation (#150). #155 is the tail — gated on all five preceding sub-issues.

```
#150 (foundation) ─┬─→ #151 (lazy + manual)  ──→
                   ├─→ #152 (partition/ceiling/strict) ──→  #155 (docs + showcases)
                   ├─→ #153 (declared predicates) ──→
                   └─→ #154 (overlay) ──→
```

Reasonable PR ordering: #150 first; #151 + #152 + #153 in parallel after; #154 after the foundation (could go in parallel with #151-#153); #155 last.

## Why this is v2

`withDerivation` v1 (#129, shipped in pre.11) handles **record-level** derivation (source record → output records). MV is the **query-level** complement: a declared query whose result is persisted, kept fresh as sources change, and read like any other collection.

Pulled into v2 during niwat-review of #149:
- `declaredDeterministicPredicates` — function-based filtering with stable hashing
[...]

Issue #133, opened 2026-05-18
fix(hub): strict-mode derivation can orphan early-written outputs on mid-output failure


vLannaAi:
## Context

D12 of the derivation plan (commit `51720cc`) fixed a related bug: a strict-mode derivation failure inside `Collection.put` now correctly rolls back the source write, because `executed.push(...)` in `runTransaction` was reordered to BEFORE the put call.

But a related multi-output edge case remains: if a derivation strategy declares N outputs and the user's `derive` function returns successfully for the first M outputs but a strict failure occurs on the (M+1)th, the first M outputs were already written to their respective collections via `Collection.put` calls in `dispatchDerivations` — those puts are NOT visible to the source transaction's `ctx._ops` array (they're nested side-effect puts, not direct `tx.put` calls).

The transaction's `revertExecuted` rolls back the source op but leaves the first M outputs orphaned on disk.

## Reproducer (illustrative)

```ts
const strategy = withDerivation({
  source: 'src',
  outputs: {
    good: { shape: 'record', collection: 'good' },
    bad:  { shape: 'record', collection: 'bad' },
  },
  derive: () => ({
    good: { value: 1 },                 // first output OK
    bad:  null as any,                  // triggers shape error (or thrown after a successful good put)
  }),
  strict: true,
  lifecycle: 'eager',
})

await db.transaction(async tx => {
  tx.vault('demo').collection('src').put('id', { ... })
})
// → source 'src/id' correctly rolled back
// → BUT 'good/id' is left on disk, orphaned (no matching source)
```

## Possible fixes

1. **Batch derived writes in the source's revert plan.** Before each derived `put`, push a synthetic op into `ctx._ops` (or a parallel "derived-ops" list) so `revertExecuted` walks them in reverse. Cleanest.

2. **All-or-nothing executor.** `DerivationExecutor.run` could buffer all output values and only call `getCollection().put` after all outputs validated. Doesn't help if `put` itself throws.

3. **Strict mode aborts the cycle BEFORE any output is written.** Run the entire derive first, validate all outputs, then write — but option 2 already covers this.

Option 1 is the most robust because it correctly composes with multi-strategy + multi-vault transactions.

## Acceptance

- Test reproducing the orphan (write strategy with 2 outputs, second one throws strict)
- After rollback, no derived records remain on disk (neither output collection has an entry for the source id)
- Existing strict-mode rollback test (D12) still passes

## References

- Branch `feat/derivation` (commit `51720cc`)
[...]

PR #149, opened 2026-05-20
docs(spec): Dim 14 v2 design — withMaterializedView (#142)

### LINKED ISSUE #142: docs(spec): Dim 14 v2 design — withMaterializedView
## Overview

Author the v2 implementation design doc for `withMaterializedView`, mirroring the structure of [`docs/superpowers/specs/2026-05-01-dim14-derivation-v1-design.md`](../blob/main/docs/superpowers/specs/2026-05-01-dim14-derivation-v1-design.md) (which graduated the v1 `withDerivation` brainstorm into a shippable spec).

This spec gates the implementation epic — sub-issues for the implementation epic should not be created until this spec lands.

## Why a spec first

`withDerivation` v1 succeeded as a single coherent epic (#129) because the design doc had already settled every architectural question. We want the same property for v2: by the time the implementation epic is broken into sub-issues, the components, invariants, and v2-vs-v3 boundary are all decided.

## Source material

- [`docs/superpowers/specs/2026-05-01-dimensions/14-derived-data.md`](../blob/main/docs/superpowers/specs/2026-05-01-dimensions/14-derived-data.md) — dimension brainstorm, including the MV API sketch at L60-69
- [`docs/superpowers/specs/2026-05-01-dim14-derivation-v1-design.md`](../blob/main/docs/superpowers/specs/2026-05-01-dim14-derivation-v1-design.md) — the v1 design doc to mirror (L43: explicitly defers MV to v2 with reason "Different shape — query → result")
- [`docs/subsystems/derivations.md`](../blob/main/docs/subsystems/derivations.md) — v1 subsystem doc
- `packages/hub/src/derivations/` — v1 implementation to extend (registry, executor, strategy-hash, stale)
- `packages/hub/src/query/` — query AST the MV will analyze (join.ts, live.ts, where/groupBy/aggregate)

## Required sections (mirroring v1 design doc)

- [ ] **Goal** — one-paragraph statement
- [ ] **Success criteria (acceptance)** — bullet list, testable
- [ ] **v2 SCOPE — what's in** — table with rationale per row
- [ ] **v2 SCOPE — what's deferred** — table with target version and reason per row
- [ ] **Architecture** — layer diagram, components, file inventory under `packages/hub/src/materialized-views/` (or extend `packages/hub/src/derivations/`)
- [ ] **Key invariants** — zero-knowledge preserved, envelope format, no new store interface, atomic with `withTransactions`
- [ ] **API surface** — exact factory signature, option types, errors
- [ ] **Query dependency analysis** — algorithm for walking a query AST → set of source collections (must handle `where`, `join`, `groupBy`, `aggregate`; note that `join('clientId')` makes `clients` a trigger source too)
- [ ] **Lifecycle** — eager (re-materialize on source-write), lazy (mark stale, materialize on read), manual (`vault.refreshView(name)`)
- [ ] **Cycle detection** — graph extends `DerivationRegistry`'s; MV → MV chains, MV → derivation chains
- [ ] **Envelope metadata** — extending `_derivedFrom` for MV (source = query hash, mvName, queryFingerprint)
- [ ] **Strict-mode rollback** — composition with `withTransactions`
- [ ] **Materialized-collection shape** — does the MV write to a regular collection, a virtual one, or a reserved namespace? How are reads routed?

## Scope questions the spec MUST resolve

1. **Scheduled refresh** (`refresh: { every: '1h' }` in the dimension brainstorm at L66) — there's no general cron/scheduler primitive in `@noy-db/hub` today (only `SyncScheduler` for sync; the rest is consumer-scheduled). Recommend deferring this to v3 alongside the Dim 11 hooks/triggers primitive. If kept in v2, the spec must define how the tick is driven.
2. **Streaming MVs** — already deferred to v3 per the v1 design doc (pairs with Dim 12 stream primitive). Spec should reaffirm and explain the boundary.
3. **MapReduce views** (CouchDB lineage, dimension brainstorm L73) — spec should explicitly state v2 vs v3.
4. **Rendered views** (Zero / Replicache lineage, dimension brainstorm L75) — spec should explicitly state v2 vs v3.
5. **`refresh: 'eager'` cost ceiling** — analogous to `JoinTooLargeError`/`GroupCardinalityError` from the query DSL. Does the MV materialization have a row ceiling? What error?
6. **Same-DEK vs separate-DEK** — v1 uses same-DEK by default. Inherit, or revisit?

## Deliverable

`docs/superpowers/specs/<date>-dim14-mv-v2-design.md` following the v1 design doc's section structure.

## Relationship to the epic

Once this spec lands and is reviewed, the implementation epic (separate issue, also in this milestone) will be split into sub-issues mapped to the spec's §Components table — same way #129 implicitly mapped to v1 §Components.

vLannaAi:
## Summary

Implementation spec for \`withMaterializedView\` — Dim 14 v2's query-level complement to v1's record-level \`withDerivation\` (shipped pre.11). Closes #142.

Mirrors the structure of [\`2026-05-01-dim14-derivation-v1-design.md\`](https://github.com/vLannaAi/noy-db/blob/main/docs/superpowers/specs/2026-05-01-dim14-derivation-v1-design.md) — same section ordering, same architecture-then-data-flow rhythm. 483 lines; 31 sections.

## Scope locks

Every scope question the issue body raised + every input the niwat consumer left is resolved in the doc, not punted:

- **MV is read-only projection.** Operator-editable lifecycle (niwat's PND.1 \`withOverlayedView\` use case) is a separate primitive, explicitly deferred to v2.5. Keeps the MV contract simple and well-defined.
- **Scheduled refresh / Streaming MVs / MapReduce / Rendered views** — all deferred to v3 with reasons stated.
- **Same-DEK encryption** — inherits v1 default; \`_materializedFrom\` payload metadata follows the \`_derivedFrom\` precedent.
- **Cost ceiling** — \`MaterializedViewTooLargeError\` at 100k rows (overridable per-MV), matching \`JoinTooLargeError\` / \`GroupCardinalityError\` shapes.
- **Empty groups** — \`onEmpty: 'delete'\` default (matches niwat's \"no payroll → no PND.1 disbursement\" expectation); \`'keep'\` opt-out.
- **Same-collection-as-source MV** (niwat DERIV-PP30-001 shape) — \`output.partition: { field, value }\` discriminator + narrow cycle-detector check that requires a provably-disjoint where-clause on the partition field.

## Architecture reuse

The spec leans heavily on primitives shipped in the recently-merged PR #148:

- **\`Collection._internalDelete\`** — refresh-driven tombstones bypass user \`onDelete\` guards (the #144/#145 composition fix). Lets MV tombstone rows on an output collection that has \`onDelete: throw\` registered.
- **Amendment-collectChange forwarding** — MV tombstones fired inside admin amendments DO surface to \`amendment.invariant\` (the \"truly unconditional\" claim correctness landed in 4d410cc).
- **\`DerivationRegistry\` cycle detection** — extended with a new node type so MV ↔ MV / MV ↔ derivation chains use a single DFS.
- **\`ReadOnlyVaultFacade.query()\`** — MV executor uses it to read source collections (the #146 surface).
- **TxContext \`_executed\` tracking** — refresh writes/tombstones register through the same #133 mechanism, atomic with \`withTransactions\`.

The implementation epic (#143) inherits all these without re-design.

## Implementation sequencing

§ \"Sequencing for implementation\" enumerates 13 steps from type-surface-only-no-execution through showcases + \`features.yaml\` entry. Pairs cleanly with how #143 is currently structured.

## Test plan

- [ ] Niwat-consumer review pass for spec-coverage / scope completeness
- [ ] Cross-check the API surface against the v1 design doc for tonal consistency
- [ ] Once approved, gate the #143 implementation epic to start; create sub-issues from § Sequencing for implementation

### DIFF
--- a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md
+++ b/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md
@@ -35,13 +35,16 @@ Ship a `withMaterializedView` strategy in `@noy-db/hub` that lets a vault declar
 | `_materializedFrom` envelope metadata (query fingerprint + queryHash + version + ts) | ✓ | Extends `_derivedFrom`; lives in encrypted payload |
 | `MaterializedViewTooLargeError` row ceiling at 100k by default | ✓ | Mirrors `JoinTooLargeError`; overridable via `{ maxRows }` |
 | Same-collection-as-source MV with `outputPartition` discriminator | ✓ | Resolves cycle detector for niwat's `DERIV-PP30-001` shape |
-| Showcase + recipe + subsystem doc + `features.yaml` entry | ✓ | One showcase per refresh mode |
+| `declaredDeterministicPredicates` on MV strategy | ✓ | Pulled forward from v2.x (niwat-review of #149). Function-based filtering with a consumer-stable `predicateHash` folded into `queryHash`. Closes the general function-predicate gap; SSO-specific cross-product is still v3. |
+| `withOverlayedView` (read-shadow variant) | ✓ | Pulled forward from v2.5 (niwat-review of #149). Narrow primitive — single shadow predicate (`shadowField` + `shadowValue`), no arbitrary merge logic. Collapses W3-A/B/C-style consumer imperative shells into one declarative block per rule. |
+| Showcase + recipe + subsystem doc + `features.yaml` entry | ✓ | One showcase per refresh mode + one for overlay + one for predicates |
 
 ## v2 SCOPE — what's deferred
 
 | Feature | Deferred to | Why |
 |---|---|---|
-| `withOverlayedView({ base: mvName, overlayCollection, mergePolicy })` | v2.5 | Operator-editable lifecycle composition (niwat PND.1 use case). Separate primitive — MV stays pure read-only projection; overlay is the merge layer on top. Designing the merge contract is a non-trivial spec on its own. |
+| `withOverlayedView` with arbitrary `mergePolicy` callback | v3 | The read-shadow variant lands in v2 (see "in" table above). Arbitrary user-supplied merge functions — priority lattices, field-level merge, history-aware reconciliation — still need a dedicated spec and stay deferred. |
+| MV cross-join (`periods × workers` style cartesian semantics) | v3 | DERIV-SSO-001's full structural path needs cross-product semantics in the query DSL — separate primitive, not query-DSL-extension territory. The v2 path for SSO is a consumer-maintained `(workerId, period)` junction collection + MV aggregate (see § DERIV-SSO-001 v2 path). |
 | Scheduled refresh (`{ every: '1h' }`) | v3 | No general cron/scheduler primitive in `@noy-db/hub`; pairs with Dim 11 hooks/triggers |
 | Streaming MVs (incremental over Dim 12 streams) | v3 | Pairs with Dim 12 stream primitive's own v1 |
 | MapReduce views (`map` + optional `reduce`, CouchDB lineage) | v3 | Different shape; separate spec |
@@ -99,9 +102,13 @@ Ship a `withMaterializedView` strategy in `@noy-db/hub` that lets a vault declar
 | `_materializedFrom` envelope ext | `packages/hub/src/envelope.ts` (modify) | New optional metadata field inside `_data` |
 | `vault.refreshView(name)` method | `packages/hub/src/vault.ts` (modify) | Bulk re-materialize entrypoint |
 | `MaterializedViewTooLargeError` | `packages/hub/src/errors.ts` (modify) | Thrown when materialization exceeds `maxRows` |
-| Showcases | `showcases/src/81-with-mv-eager.showcase.test.ts`, `82-with-mv-lazy.showcase.test.ts` | One per refresh mode |
-| Subsystem doc | `docs/subsystems/derivations.md` (extend) | New § Materialized Views section |
-| `features.yaml` entry | `features.yaml` (modify) | New `materialized-views` row under the derivation cluster |
+| `withOverlayedView()` factory | `packages/hub/src/overlay-views/with-overlayed-view.ts` | API surface for the read-shadow overlay primitive |
+| `OverlayedViewRegistry` | `packages/hub/src/overlay-views/registry.ts` | Name→{base, overlay, shadow} mapping; shared cycle detection |
+| `OverlayedCollection` proxy | `packages/hub/src/overlay-views/virtual-collection.ts` | `Collection<T>`-shaped merge-on-read + write-to-overlay routing |
+| `QueryBuilder.wherePredicate(name, ctx?)` | `packages/hub/src/query/builder.ts` (modify) | Declared-predicate terminal added to the chainable builder |
+| Showcases | `showcases/src/81-with-mv-eager.showcase.test.ts`, `82-with-mv-lazy.showcase.test.ts`, `83-with-overlay.showcase.test.ts`, `84-with-mv-predicates.showcase.test.ts` | One per refresh mode + overlay + predicates |
+| Subsystem doc | `docs/subsystems/derivations.md` (extend) | New § Materialized Views + § Overlay views sections |
+| `features.yaml` entry | `features.yaml` (modify) | New `materialized-views` + `overlay-views` rows under the derivation cluster |
 
 ### Modified components
 
@@ -207,12 +214,91 @@ interface MaterializedViewStrategy<TRow> {
    * when the domain warrants it.
    */
   maxRows?: number
+  /**
+   * Declared deterministic predicates that the query may invoke via
+   * `.wherePredicate(name, ctx?)`. Function bodies have no stable
+   * canonical serialization, so the consumer supplies a stable
+   * `hash` per predicate — included in `queryHash` so a change to
+   * the function's semantics forces a refresh on next visit (mirrors
+   * v1's `strategyHash` mechanism for `derive`).
+   *
+   * Use when a `.where(field, op, value)` clause is too narrow:
+   * boolean predicates over array fields, date-range membership,
+   * cross-field invariants. The predicate fn receives the source row
+   * + an optional `ctx` argument the query call site supplies.
+   *
+   * Consumer responsibility: bump `hash` whenever the function's
+   * semantics change. Stale hashes do not corrupt; they just trigger
+   * a re-materialization that may not actually be needed.
+   */
+  predicates?: {
+    [name: string]: {
+      hash: string
+      fn: (row: unknown, ctx?: unknown) => boolean
+    }
+  }
 }
 
 // Returned by withMaterializedView()
 interface MaterializedViewStrategyHandle {
   __noydb_strategy: 'materialized-view'
   spec: MaterializedViewStrategy<unknown>
+}
+
+// New: overlay primitive — read-shadow merge of two collections
+interface OverlayedViewStrategy {
+  /**
+   * Virtual collection name. `vault.collection(name)` returns a view
+   * that merges `base` and `overlay` per the shadow rule. Reads union
+   * by id; writes route to `overlay` only (the `base` is owned by an
+   * MV or another upstream).
+   */
+  name: string
+  /**
+   * The collection providing the default rows. Typically an MV's
+   * output collection. The overlay primitive does NOT modify `base`
+   * — it only shadows on read.
+   */
+  base: string
+  /**
+   * User-writable collection that carries overrides. Independent
+   * write path; supports the standard `Collection` API including
+   * `withGuard` / `withDerivation` registration.
+   */
+  overlay: string
+  /**
+   * Single-field shadow predicate. When `overlay[shadowField] ===
+   * shadowValue` for a given id, reads of that id return the overlay
+   * row; otherwise reads return the base row. Designed to support
+   * "the operator can flip a row to 'override' mode and from then on
+   * their hand-edited values win." Niwat's `dataStatus === 'override'`
+   * is the canonical example.
+   *
+   * No callback merge, no priority lattice, no field-level merge —
+   * v2 stays explicitly narrow. Arbitrary merge policies are v3.
+   */
+  shadowField: string
+  shadowValue: unknown
+}
+
+interface OverlayedViewStrategyHandle {
+  __noydb_strategy: 'overlayed-view'
+  spec: OverlayedViewStrategy
+}
+
+// Query DSL extension (added to QueryBuilder for MV consumers)
+interface QueryBuilder<T> {
+  /**
+   * Invoke a declared predicate by name. The predicate must be
+   * registered on the calling MV's `predicates` map; `queryHash`
+   * folds the predicate's stable `hash` so a function-body change
+   * (via `hash` bump) is detected at refresh time.
+   *
+   * `ctx` is opaque to the query DSL — passed verbatim to the
+   * predicate fn. Useful for threading config values into the
+   * predicate (date ranges, feature flags, etc.).
+   */
+  wherePredicate(name: string, ctx?: unknown): QueryBuilder<T>
 }
 
 // Vault method (added)
@@ -290,22 +376,54 @@ The set is materialized at MV registration time. `MaterializedViewRegistry.onSou
 - The MV output collection must not overlap with the source set, UNLESS `output.partition` is declared AND the input query has a `.where(partition.field, '!=', partition.value)` clause (or equivalent — see § Cycle detection below).
 - The query plan must be deterministic — no `.subscribe()` / `.live()` terminals embedded (these would create a re-entrant cycle).
 
-### Function-based source-row predicates are not supported in v2
+### Function-based source-row predicates (`declaredDeterministicPredicates`)
 
-The MV `query()` builder accepts the standard `.where(field, op, value)` terminals + `.join` / `.groupBy` / `.aggregate`. Function-based predicates (e.g. niwat's `DERIV-SSO-001` "active employees during the period" rule, where the active-ness check is a function over an `employmentPeriods` array on the worker record) are **not supported** in v2. Reason: `queryHash` determinism requires a stable canonical serialization of the query plan; function bodies don't have one.
-
-**Workaround pattern (canonical for v2):** pre-derive the predicate result via a v1 `withDerivation`, then MV-aggregate over the pre-derived collection. For DERIV-SSO-001:
+The MV `query()` builder accepts the standard `.where(field, op, value)` terminals + `.join` / `.groupBy` / `.aggregate`. For cases where these aren't expressive enough — boolean predicates over array fields, date-range membership, cross-field invariants — v2 adds `declaredDeterministicPredicates`:
 
 ```ts
-// v1 derivation: per (worker, period) row with active-ness pre-computed
-withDerivation<Worker, { activeInPeriod: { id: string; workerId: string; period: string; active: boolean } }>({
-  source: 'workers',
-  outputs: { activeInPeriod: { shape: 'record', collection: 'worker-period-active' } },
-  derive: (w, ctx) => ({ /* per-period explosion using w.employmentPeriods */ }),
-  lifecycle: 'eager',
+withMaterializedView({
+  name: 'overdue-aggregate',
+  predicates: {
+    isOverdue: {
+      hash: 'is-overdue-v1',
+      fn: (inv: Invoice, ctx: { asOf: string }) =>
+        inv.status === 'open' && inv.dueDate < ctx.asOf,
+    },
+  },
+  query: () => invoices.query()
+    .wherePredicate('isOverdue', { asOf: '2026-05-20' })
+    .groupBy('clientId')
+    .aggregate({ outstanding: sum('amount') }),
+  rowKey: (r) => r.clientId,
+  refresh: 'eager',
 })
+```
 
-// v2 MV: aggregates over the pre-derived collection — pure where-clause query
+The function body's lack of canonical serialization is handled by the consumer-supplied `hash` field — folded into `queryHash` so a function change (signalled by a `hash` bump) forces a refresh on next visit. Same drift-detection pattern v1 `withDerivation` uses for `derive.toString()`.
+
+**Consumer responsibility:** bump `hash` whenever the predicate's semantics change. Failing to bump is not unsafe (the old hash matches the old MV rows; new rows just use the new function); failing to bump after a *non-equivalent* function change leaves stale rows around until the next refresh.
+
+#### DERIV-SSO-001 v2 path
+
+DERIV-SSO-001 from the niwat enforcement plan reads workers filtered to "active during the period" and counts per period. This needs **cross-product** semantics (periods × workers), which v2's query DSL does NOT add — predicate filtering operates on a single row at a time, and v2 has no cartesian join (deferred to v3).
+
+The honest v2 path for SSO is a **consumer-maintained junction collection** combined with a predicate filter. The junction is small enough (~niwat scale: 12 periods × 30 workers = 360 max rows) to maintain from app code without losing too much:
+
+```ts
+// niwat-app: junction collection. Rebuilt when a Worker's
+// employmentPeriods edit fires. ~30 LoC in a Vue composable.
+type WorkerPeriodActive = {
+  id: string            // `${workerId}|${period}`
+  workerId: string
+  period: string
+  active: boolean       // computed at write time from worker.employmentPeriods
+}
+
+// v2 MV: aggregates the junction by period. Predicate not strictly
+// needed here — `where('active', '==', true)` would do — but if
+// the consumer wants to derive `active` from the row instead of
+// storing it, the predicate primitive enables that without a junction
+// rewrite.
 withMaterializedView({
   name: 'sso-aggregate',
   query: () => workerPeriodActive.query()
@@ -317,7 +435,9 @@ withMaterializedView({
 })
 ```
 
-The composition is intentional — v1 owns "compute a deterministic field per source row," v2 owns "aggregate/project over rows by declared predicate." A future v2.x may add `declaredDeterministicPredicates: { activeInPeriod: fn }` with a registered `predicateHash` so the function can live inside the MV query directly, but v2 keeps the surface narrow.
+Cost vs the v3 cross-join path: ~30 LoC of junction-maintenance code in the consumer, repeated for each per-period rule. Smaller than the imperative shell `withOverlayedView` replaces; structural enforcement still holds *within* the predicate (the consumer can't accidentally count an inactive worker) but the junction maintenance itself is application-owned.
+
+A future v3 cross-join primitive (`workers.query().crossJoin('periods', { as: 'p' }).wherePredicate('activeInPeriod', { period: '$row.p.id' })`) closes the remaining structural gap. Not in v2.
 
 ## Lifecycle
 
@@ -380,11 +500,12 @@ Same MaterializedViewExecutor.refresh path; no implicit triggers.
 
 Single shared graph with v1 derivations:
 
-- Node types: `'derivation'` (v1) and `'materialized-view'` (v2). Each node is keyed by its `source` (derivation) or `name` (MV).
+- Node types: `'derivation'` (v1), `'materialized-view'` (v2), `'overlayed-view'` (v2). Each node is keyed by its `source` (derivation) or `name` (MV / overlay).
 - Edges:
   - Derivation: source collection → output collections (existing v1)
   - MV: every collection in the dependency set → MV's output collection (new)
-- Detection: DFS from each registered node at vault open. Throws `DerivationCycleError` (existing) or `MaterializedViewCycleError` (new) on any back-edge. Same vault-init failure mode as v1.
+  - Overlay: `base` → `name` and `overlay` → `name` (new)
+- Detection: DFS from each registered node at vault open. Throws `DerivationCycleError` (existing), `MaterializedViewCycleError` (new), or `OverlayedViewCycleError` (new) on any back-edge. Same vault-init failure mode as v1.
 
 ### Same-collection-as-source MV (partition discriminator)
 
@@ -399,6 +520,7 @@ withMaterializedView({
     .where('type', 'in', ['vatSales', 'vatPurchase', 'vatCredit'])
     .groupBy('period')
     .aggregate({ net: sum('amount') }),
+  rowKey: (r) => `pp30|${r.period}`,
   output: {
     collection: 'disbursements',
     partition: { field: 'type', value: 'pp30' },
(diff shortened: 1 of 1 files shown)


vLannaAi:
**Niwat-consumer review pass.** Walked the 31-section spec against the niwat enforcement plan (W3 derivations + W4 freezes + adoption blockers). Spec covers every consumer use case I raised on #142 / #144 / #145 + everything that came out of PR #148 review. Going section-by-section against the niwat enforcement plan:

### Niwat use-case coverage

| Rule | Spec section | Niwat shape | Verdict |
|---|---|---|---|
| `DERIV-PND1-001` (compensations Σ taxAmount) | § Lifecycle, § Type surface | `withMaterializedView({ name: 'pnd1-aggregate', query: () => compensations.query().groupBy(['clientId', 'period']).aggregate({ taxTotal: sum('taxAmount') }), refresh: 'eager' })` | ✅ Direct fit. `onEmpty: 'delete'` matches "no payroll → no PND.1 disbursement." |
| `DERIV-PP30-001` (same-collection partition) | § Same-collection-as-source MV | Source filter `.where('type', 'in', ['vatSales', 'vatPurchase', 'vatCredit'])` is in the listed disjoint-provability set; output partition `{ field: 'type', value: 'pp30' }` is verbatim the worked example in the spec | ✅ Spec accommodates exactly the niwat shape |
| `DERIV-PAY-ALLOC → RECEIPT` (1:1) | n/a (this is v1 `withDerivation`, not MV) | Already shipped via PR #148 with optional outputs (#144) | ✅ Out of scope for v2 — confirmed |
| `RCT-CANCEL-001` + MV tombstones | § Key invariants, § Error handling | `_internalDelete` bypass docs `onDelete` won't trip; admin amendment cascade visible to `amendment.invariant` | ✅ Resolution from PR #148 carried forward correctly |
| `DERIV-SSO-001` (workers active in period) | § Query dependency analysis | **See concern below** | ⚠ Concern (1) |

### Concern (1) — function-based predicates: `DERIV-SSO-001`'s active-employment filter

The one real adoption blocker I see. `DERIV-SSO-001` per the niwat rulebook reads `workers` filtered to "active employees during the period," where the predicate is `wasEmployedDuring(periods, period)` — a function over `employmentPeriods` array. The MV `query()` builder only accepts `.where(field, op, value)` terminals, and `queryHash` determinism (open question #3) requires a stable serialization that function-based predicates don't have today.

Three resolutions, in order of preference:

1. **Pre-derive an `isActiveInPeriod` boolean per (worker, period) via v1 `withDerivation`**, then the MV groupBy's over that pre-derived collection. Heavier (extra collection), but it composes existing primitives — no spec change. Probably what niwat ships with v2.
2. **Extend the spec/query DSL to allow declared deterministic predicates** with a `predicateHash` registered alongside the MV. Could be v2.x. Best long-term answer; adds non-trivial surface.
3. **Defer DERIV-SSO-001 entirely** until v2.5 / `withOverlayedView` lands. Niwat's enforcement plan already had SSO in W3-B (after PND.1), so this is acceptable.

Niwat prefers (1) for v2 since it requires no spec change. Worth noting in § Open implementation questions that "function-based source-row filters need pre-derivation via v1 `withDerivation`" so future consumers know.

### Concern (2) — operator-editable lifecycle deferral consequences

The spec correctly defers `withOverlayedView` (operator-editable lifecycle on top of MV) to v2.5. Consequence for niwat: **W3-A/B/C don't become fully structural in pre.14**. Niwat's `Disbursement` carries `dataStatus` / `filingStatus` / `overrideAt` / etc. — an MV alone can't own these. The actual W3-A shape niwat ships in pre.14:

- New collection `pnd1-aggregate` owned by the MV (per § Materialized-collection shape)
- `disbursements` stays imperative; `upsertWhtS01Disbursement` reads `pnd1-aggregate` for the amount, owns the lifecycle, applies `DSB-OVERRIDE-001` short-circuit
- When `withOverlayedView` ships in v2.5, the two collections collapse into one structural primitive

Smaller cost than the shadow-fragment pattern I sketched in `withDerivation-sketches.md` Option A (no double-storage of `taxAmount`), but the imperative shell stays. Worth a sentence in the spec's § Backward compatibility or a new § Composition with operator lifecycle saying "MV + imperative shell is the canonical v2 pattern until v2.5 ships overlay." Helps future consumers understand the boundary.

**Any tentative ETA on v2.5?** Doesn't need to be on this PR but informs niwat's W3 rollout cadence.

### Small clarifications

- **`output.collection` defaulting to `name`** (§ Type surface line 159): if a consumer registers `withMaterializedView({ name: 'pnd1-aggregate', ... })` without `output.collection`, the MV writes to `pnd1-aggregate`. But that collection doesn't pre-exist in the schemas list — does noy-db auto-register it, or does the consumer need to declare an empty schema entry first? Worth one line in the spec.
- **`rowKey` for groupBy MVs uses tuple `${clientId}/${period}`** (open question #6): niwat's keys are safe (`c1`, `2026-05` — no slashes), but a key containing `/` would silently collide. Recommend documenting the separator as part of the convention OR requiring `rowKey` always (groupBy too). Minor.
- **`refresh: 'eager'` cost ceiling at 100k default** (§ Error handling): niwat's largest aggregate is ~3,600 rows (30 clients × 12 months × ~10 workers). 100k is fine. Override per-MV is the right escape hatch.
- **Cross-collection joins inside MV** (§ Query dependency analysis): the analyzer follows `.join('foreignKey', { as: 'alias' })`. Niwat's PND.53 future use-case might join `paymentAllocations` with `clients` to slice by counterparty. As long as the joined collection's writes also trigger refresh, this works. Confirmed by the dep-set semantics.

### Open-question feedback

- (#1, **diff-strategy for incremental refresh**) — Niwat doesn't need incremental for v2; full re-materialize is fine at our scale. v2.x optimization is the right placement.
- (#2, **refresh ordering**) — Sequential FIFO matches niwat's mental model; we won't have order-dependent MV chains in v2. If a future consumer needs ordering, opt-in topological is the natural extension.
- (#3, **`queryHash` determinism**) — Tied to Concern (1) above. If the spec adopts a future "declared deterministic predicates" path, the predicate's hash needs to fold in. v2 default of "no function predicates" is fine.
- (#6, **`rowKey` default**) — Lean toward "require `rowKey` for all MVs" (groupBy too) — eliminates the slash-collision class. Minor ergonomic cost; structural clarity gain.

### Net verdict

Spec is consumer-ready for v2's stated scope. Concerns (1) and (2) don't change the spec's shape, they're observations about what v2 enables vs what waits for v2.5. The "MV is read-only projection" stance is the right architectural call — keeps v2 simple and bounds the merge-policy spec for v2.5 cleanly.

niwat is happy to be the test consumer for the #143 implementation epic once this lands. Specifically the same-collection partition + `_internalDelete` tombstone interaction is the place we'd push hardest in the integration showcase.

vLannaAi:
**Resolved in `2741520`** (pushed to the spec branch).

Four targeted responses to the niwat-review pass, ordered by impact:

### 1. `rowKey` required for all MVs (Small clarification #2 + open question #6)

Moved from "open question with proposed default" to "resolved decision: explicit always." Added `rowKey: (row: TRow) => string` to `MaterializedViewStrategy` — no default. Eliminates the `/`-collision class for groupBy keys; minor ergonomic cost; structural clarity gain. Niwat's lean ("require `rowKey` always") wins.

### 2. Function-based source predicates explicitly unsupported in v2 (Concern 1)

Added a new § "Function-based source-row predicates are not supported in v2" under § Query dependency analysis, documenting the canonical workaround: pre-derive via v1 `withDerivation`, MV-aggregate over the pre-derived collection. Worked example using `DERIV-SSO-001` (per (worker, period) active-ness pre-computed by a v1 derivation; v2 MV aggregates the result). `declaredDeterministicPredicates` stays open as a future v2.x extension. Moved from open questions to § Resolved during niwat-review.

### 3. Output-collection accessibility clarified (Small clarification #1)

One-line clarification in § Materialized-collection shape: noy-db collections are dynamically accessible via `vault.collection(name)` — no `schemas` pre-declaration. The MV registration itself is the binding.

### 4. § "Composition with operator-editable lifecycle (v2.5 bridge)" (Concern 2)

New section between § Materialized-collection shape and § Error handling. Documents the canonical MV + imperative shell pattern for v2:

- MV writes pure computed values to its own collection
- Separate user-owned collection carries the operator lifecycle (dataStatus, filingStatus, overrideAt, ...)
- Worked example: `upsertWhtS01Disbursement` reading from `pnd1-aggregate`, applying `DSB-OVERRIDE-001` short-circuit

Documented v2.5 ETA: **TBD** — depends on pre.14 close + real-consumer demand assessment after MV ships. Honest about the bridge's cost (two collections, imperative upsert, no automatic re-upsert on re-materialization) and what v2.5 collapses (single structural primitive, declarative upsert, lifecycle fields in the overlay declaration).

### Open-question feedback acknowledged in place

- (#1 incremental refresh) — confirmed v2.x as the right placement
- (#2 refresh ordering) — confirmed sequential FIFO is right
- (#3 queryHash determinism) — added cross-reference to the future predicate path

### Verdict on spec readiness

Niwat's net verdict ("Spec is consumer-ready for v2's stated scope") stands — these are sharpening edits, not shape changes. Concerns 1 and 2 are now explicit constraints rather than implicit assumptions. The four small clarifications are folded in or addressed by the `rowKey`-required decision.

Length: 483 → 566 (+83 lines, +17%). Mostly the v2.5 bridge section and the function-predicate workaround example.

Happy to take another look from niwat against the niwat enforcement plan one more time before merge, or merge-and-iterate as part of the #143 implementation epic.

vLannaAi:
### Code review

Found 4 issues — all internal spec inconsistencies introduced when commit `2741520` made `rowKey` required:

1. `pp30-aggregate` worked example missing required `rowKey` field. The example is the canonical pattern for same-collection-partition MVs (niwat's DERIV-PP30-001), but lacks the now-required `rowKey` declared in § Type surface.

https://github.com/vLannaAi/noy-db/blob/2741520da0cb309c1404310ceb4cc15902bf3a42/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L394-L408

2. § "Composition with operator-editable lifecycle (v2.5 bridge)" — the `upsertWhtS01Disbursement` consumer code reads from `pnd1-aggregate` via `${clientId}|${period}`, but the `pnd1-aggregate` MV registration is never shown anywhere in the spec. The `rowKey` that produces this key is implied but missing — readers must infer the MV declaration. Should be made self-contained.

https://github.com/vLannaAi/noy-db/blob/2741520da0cb309c1404310ceb4cc15902bf3a42/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L451-L466

3. DERIV-SSO-001 workaround is structurally impossible in v1. The `withDerivation` block uses `outputs: { activeInPeriod: { shape: 'record', collection: 'worker-period-active' } }` with the comment "per-period explosion using `w.employmentPeriods`." But v1 `withDerivation` only supports `shape: 'record'` (1:1 source→output), per [`derivations/types.ts`](https://github.com/vLannaAi/noy-db/blob/2741520da0cb309c1404310ceb4cc15902bf3a42/packages/hub/src/derivations/types.ts#L24-L31). A single Worker can only emit ONE `activeInPeriod` row, not N. The proposed MV `.groupBy('period')` then has nothing to aggregate. The workaround as written cannot work with current primitives. Either v1 needs multi-row outputs (a v2.x feature) or the spec needs a different canonical pattern.

https://github.com/vLannaAi/noy-db/blob/2741520da0cb309c1404310ceb4cc15902bf3a42/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L298-L320

4. § Integration tests showcase description omits `rowKey`. The PND.1 showcase reads "writes to `pnd1-aggregate`" with no `rowKey` shown — incomplete test spec since `rowKey` is now required. Same shape as issue 1.

https://github.com/vLannaAi/noy-db/blob/2741520da0cb309c1404310ceb4cc15902bf3a42/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L501-L510

Issue 3 is the load-bearing one — the proposed canonical workaround pattern for "function-based predicates" doesn't actually work, which means DERIV-SSO-001 has no clean v2 path without either pulling forward `declaredDeterministicPredicates` or adding multi-row v1 outputs.

vLannaAi:
**Niwat-consumer push: anticipating deferred items into pre.14.**

Looking at the spec's deferral list against niwat's W3-W4 adoption plan, two of the deferred items would meaningfully simplify the niwat app if pulled forward into this milestone. Posting as a consumer-side argument, with no expectation either lands as-is — but the math matters for sequencing.

### 1. `withOverlayedView` (v2.5 → v2) — at minimum a "read-shadow" variant

**The bridge pattern in § Composition with operator-editable lifecycle requires throwaway imperative code.** The `upsertWhtS01Disbursement` example in the spec is exactly what niwat would ship in pre.14 — and exactly what gets rewritten when v2.5 ships overlay. The throwaway code surface is:

- W3-A (DERIV-PND1-001): one `upsertPnd1FromAggregate` function
- W3-B (DERIV-SSO-001): one `upsertSsoFromAggregate` function (assuming Issue 3 is resolved)
- W3-C (DERIV-PP30-001): one `upsertPp30FromAggregate` function
- Future disbursement types (PND.3, PND.53 expansion, PP.36): each one repeats the pattern

That's 3+ instances of effectively identical imperative shell code that the spec says will be replaced. Real consumer cost in code-to-throw-away terms.

**The minimum-useful overlay primitive niwat would need is small.** Read-shadowing only, no arbitrary merge:

```ts
withOverlayedView({
  base: 'pnd1-aggregate',          // MV collection
  overlay: 'pnd1-overlay',         // user-owned writes
  shadowField: 'dataStatus',       // when overlay.shadowField === 'override', overlay row wins
  shadowValue: 'override',
})
```

Semantics:
- Reads from `vault.collection('pnd1')` (a virtual collection) merge: if overlay row exists AND `overlay.dataStatus === 'override'`, return overlay row; else return the MV row
- Writes to `vault.collection('pnd1')` route to the overlay collection only
- The MV refresh continues to own `pnd1-aggregate` end-to-end; the overlay collection is independently mutable

This avoids the "arbitrary merge policy" spec complexity (which justifies the v2.5 deferral) while delivering the structural primitive niwat needs. Three field declarators (`base`, `overlay`, `shadowField` + `shadowValue`) — no callback merge logic, no priority lattice, just a single shadowing predicate.

**If this is too much for v2:** even a *narrower* primitive that just returns "overlay row OR MV row" with NO merge logic at all would let niwat ship W3-A/B/C as one declarative block per rule instead of three imperative wrappers. The operator-editable fields then live entirely on the overlay collection; the MV stays pure projection.

**Scope estimate:** roughly the size of `withDerivation` v1 (~300-500 LoC including tests + dep tracking). Strong candidate for pre.14 sub-issue.

### 2. `declaredDeterministicPredicates` (v2.x → v2) — forced by Issue 3

The Issue 3 finding (DERIV-SSO-001 workaround is structurally impossible in v1) makes this no longer optional. If the workaround in the spec doesn't actually work, niwat has three fallback paths for SSO in v2:

| Path | Cost |
|---|---|
| A. Imperative shadow `worker-period-active` collection maintained by app code | Throws away the structural-enforcement promise for DERIV-SSO-001; ~150 LoC of imperative bookkeeping in niwat |
| B. Wait for `declaredDeterministicPredicates` in a future v2.x | DERIV-SSO-001 stays imperative until then; SSO is W3-B (after PND.1) so blocks but not critical |
| C. Add multi-row outputs to v1 `withDerivation` | Probably larger surface than the predicate alternative; touches a shipped subsystem |

Niwat's order of preference: include `declaredDeterministicPredicates` in v2 (option implicit-D); otherwise B. Either way, the spec's current claim that "Niwat's preferred resolution: pre-derive via v1 withDerivation, MV-aggregate over the pre-derived collection. No spec change, but document the canonical pattern" is **factually incorrect** until either v1 supports multi-row or v2 supports declared predicates.

The `declaredDeterministicPredicates` surface is small:

```ts
withMaterializedView({
  name: 'sso-aggregate',
  predicates: {
    activeInPeriod: {
      hash: 'active-in-period-v1',   // consumer-stable; changes force refresh like strategyHash
      fn: (worker, period) => worker.employmentPeriods.some(p => ...),
    },
  },
  query: () => workers.query()
    .wherePredicate('activeInPeriod', { period: '$context.period' })
    .groupBy('period')
    .aggregate({ workerCount: count() }),
  ...
})
```

The `$context.period` syntax is one approach to threading the groupBy key into the predicate; alternatives exist. Spec-wise the addition is: a new `predicates` field on `MaterializedViewStrategy`, a `.wherePredicate()` terminal on the query builder, and `predicateHash` folded into `queryHash`. ~100 LoC + tests.

### What I'm NOT pushing to anticipate

For completeness, the rest of the v3 / v2.x deferred items are correctly placed:

- Scheduled refresh `{ every: '1h' }` — niwat has no cron need
- Streaming MVs — no stream sources
- MapReduce views — query DSL aggregate is sufficient
- Rendered views / public CDN — single-vault, on-device
- Cross-vault MV sources — out of scope
- Diff-strategy incremental refresh — niwat's scale fine with full re-materialize
- `refreshView` parallel concurrency — sequential FIFO matches mental model
- MV with separate DEK — no use case

### Summary

| Pull-forward | Niwat value | Rough scope | Recommendation |
|---|---|---|---|
| `withOverlayedView` (read-shadow variant) | HIGH — collapses W3-A/B/C imperative shells | ~300-500 LoC | Strong pull-forward case; minimum-useful variant fits v2 scope |
| `declaredDeterministicPredicates` | HIGH (forced by Issue 3) | ~100 LoC | Should be in v2 — the current v2 "workaround" doesn't actually work |
| All other deferred items | LOW for niwat | n/a | Keep deferred |

If neither pulls forward, niwat's W3-A and W3-C still ship in pre.14 (PND.1 + PP30 via MV + imperative shell), but W3-B (SSO) stays imperative entirely. With both pulled forward, W3-A/B/C all become single declarative primitives.

vLannaAi:
**Resolved in `c857a6c`** — option C from the consumer-decision: pull both `declaredDeterministicPredicates` and `withOverlayedView` (read-shadow variant) forward into v2.

### Four inconsistencies (Comment 1) — fixed

- ✅ `pp30-aggregate` example now declares `rowKey: (r) => \`pp30|${r.period}\``
- ✅ v2.5 bridge example self-contained — the canonical pattern now shows both `withMaterializedView` (the `pnd1-aggregate` decl) AND `withOverlayedView` (the merge), no inferred declarations
- ✅ Integration test blurb for PND.1 explicitly mentions `rowKey: (r) => \`${r.clientId}|${r.period}\``
- ✅ **Issue 3** — the misattributed "Niwat's preferred resolution" text removed; v1-derivation pre-materialize workaround removed entirely since it doesn't work (v1 is 1:1, can't produce the junction)

### Consumer-side pulls (Comment 2) — both landed in v2

**`declaredDeterministicPredicates`** (was v2.x):
- New `predicates?: { [name]: { hash, fn } }` field on `MaterializedViewStrategy`
- New `QueryBuilder.wherePredicate(name, ctx?)` terminal
- `predicateHash` folds into `queryHash` so a `hash` bump triggers refresh on next visit
- Sequencing step 13 in § Sequencing for implementation
- Scope ≈ 100 LoC + tests

**`withOverlayedView`** (was v2.5, read-shadow variant only):
- New strategy: `{ name, base, overlay, shadowField, shadowValue }`
- Reads: overlay row wins iff `overlay[shadowField] === shadowValue`; else base
- Writes route to overlay; `vault.collection('pnd1').delete(id)` is "un-override" (removes overlay row, base resurfaces)
- Single-field shadow only — no arbitrary `mergePolicy` callback (that stays deferred to v3)
- Cycle detection edges in shared `DerivationRegistry` graph
- Sequencing steps 14-15
- Scope ≈ 300-500 LoC + tests

### DERIV-SSO-001 — honest v2 path

The predicate primitive closes the general function-based-filter gap, but SSO specifically needs cross-product (periods × workers) — v2 query DSL doesn't add cross-join, deferred to v3 (added to § what's-deferred explicitly).

The honest v2 SSO path is now documented under § DERIV-SSO-001 v2 path: a consumer-maintained `(workerId, period)` junction collection (~30 LoC in a Vue composable at niwat's scale: 12 periods × 30 workers = 360 max rows) + MV aggregate. Trade-off documented honestly: structural enforcement within the predicate, junction maintenance application-owned.

### What got reframed

The § "Composition with operator-editable lifecycle" section was a "v2.5 bridge with imperative shell" — it's now the **canonical v2 declarative pattern** (`withMaterializedView` + `withOverlayedView`). The worked example for PND.1 + DSB-OVERRIDE-001 ships as two declarative blocks, no `upsertWhtS01Disbursement` imperative shell at all.

### Length / scope

566 → 741 lines (+175). Sequencing extended from 13 to 16 steps; the two new primitives slot in at the tail (steps 13-15), independent of steps 1-12 — they can stack without forcing a tree-rewrite of the prior implementation work.

Single comprehensive commit on top of the existing branch — happy to break it into smaller commits if that helps review.

vLannaAi:
### Code review

Found 1 issue (score ≥ 80):

1. Open question #3 contains a stale cross-reference to `declaredDeterministicPredicates` as a v2.x feature, but this commit landed it in v2 (the predicates section now exists at § "Function-based source-row predicates"). The parenthetical needs to be retracted — should say "the predicate's `predicateHash` folds into `queryHash` — see § Function-based source-row predicates" without the "v2.x adds" preamble.

https://github.com/vLannaAi/noy-db/blob/c857a6cf3952830d7fdf4da1f4bbd27784f4ad1d/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L697-L702

Four additional concerns scored 75 (below merge-blocking threshold). Surfacing them in a follow-up comment below — they're real design questions about the new pull-forward surface, not cosmetic.

vLannaAi:
**Follow-up — four design questions scored 75 in re-review.** Surfacing as discussion since the maintainer pulled `declaredDeterministicPredicates` + `withOverlayedView` into v2 partly on niwat's behalf — these are the rough edges niwat would hit on adoption.

### 1. `.wherePredicate(name, ctx?)` — `ctx` not folded into `queryHash`

The `predicates: { hash, fn }` shape stably folds `predicateHash` into `queryHash`. But the `ctx` parameter passed at `.wherePredicate('isOverdue', ctx)` is **not** part of any hash and not tracked for stale invalidation. The worked example shows:

```ts
.wherePredicate('isOverdue', { asOf: '2026-05-20' })
```

…hardcoded. For the real-world use case the example is presumably illustrating ("overdue as of today") this is broken: with `refresh: 'eager'` the MV is permanently frozen to that date, and with `refresh: 'lazy'` the stale bit only fires on source-collection writes, so the date silently rolls behind.

Possible resolutions:
- Document that `ctx` must be a stable value across the MV's lifetime (and that time-dependent predicates require `refresh: 'manual'` + an external trigger)
- Fold a structural hash of `ctx` into `queryHash` at materialization time (`ctx`-change forces re-materialization on next read)
- Mark predicates-with-ctx as an explicit v2.x deferral and keep v2 ctx-less (consumer bakes everything into `hash`)

Niwat's preference is option 2 — the consumer-supplied `hash` covers function-body semantics, an auto-derived `ctx`-hash covers runtime parameters. Symmetric, no surprises.

https://github.com/vLannaAi/noy-db/blob/c857a6cf3952830d7fdf4da1f4bbd27784f4ad1d/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L380-L420

### 2. `QueryDependencyAnalyzer` overlay-awareness

`withOverlayedView` declares cycle-detection edges (`base → name`, `overlay → name`) into the shared `DerivationRegistry` graph. But the spec doesn't say what happens when a **second** MV uses the overlay's virtual collection name as its source. `QueryDependencyAnalyzer` walks `QueryPlan.rootCollection` and finds the virtual name (e.g. `pnd1` — the overlay name); there's no documented mechanism for resolving it through `OverlayedViewRegistry` to discover the concrete underlying sources (`pnd1-aggregate`, `pnd1-overlay`).

The `joinResolver.resolveRef` pattern exists for joins; the analyzer probably needs an analogous `overlayResolver.resolveVirtual` for overlay names. Otherwise: source writes to `pnd1-aggregate` or `pnd1-overlay` wouldn't trigger refresh of the downstream MV, silently serving stale data.

For niwat this isn't an immediate concern (W3-A/B/C don't currently chain MV-on-overlay), but it's the kind of thing that ships latent for a release cycle then bites.

### 3. Overlay merge semantics for overlay-only rows are undefined

The read rule is "`overlay[shadowField] === shadowValue` → return overlay; else return base." But two paths aren't covered:

- **Overlay row exists, predicate fails, no base row.** What does `get(id)` return? `null`? The orphaned overlay row? The spec is silent.
- **`vault.collection('overlay-virtual').delete(id)` when no base row exists.** The "base resurfaces" language implies base always exists. What's the post-delete state?

These are the orphaned-override cases — niwat would hit them if an operator preemptively writes an `override` row before the MV ever materializes for that key (e.g., they know the auto-computed PND.1 amount is wrong and want to lock the override in before the first compensation lands). The spec needs a row in the operations table for "overlay-only" + a sentence on the delete semantics.

https://github.com/vLannaAi/noy-db/blob/c857a6cf3952830d7fdf4da1f4bbd27784f4ad1d/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L605-L645

### 4. `withOverlayedView` requires overlay row id == MV `rowKey(baseRow)` — undocumented constraint

The PND.1 example sets `rowKey: (r) => \`${r.clientId}|${r.period}\`` on the MV. The overlay merge happens by id-match between overlay and base. So overlay rows MUST be written with `id` equal to the MV's `rowKey` output, otherwise the merge silently produces an orphan (overlay row that never shadows anything).

The spec doesn't state this constraint, validate it, or surface a helper. niwat consumer code would be:

```ts
await vault.collection('pnd1-overlay').put({
  id: `${clientId}|${period}`,   // must match MV's rowKey shape — implicit
  dataStatus: 'override',
  amount: 99999,
})
```

…and an operator who fat-fingers the separator gets a silent orphan. Recommended fix shapes:

- Surface the MV's `rowKey` as a helper on the overlay registration: `overlay.idFromBase(baseRow)` so the consumer doesn't reinvent the shape
- OR document the constraint explicitly with a code example showing the matching id format
- OR validate at write time (overlay schema check warns when the id doesn't match the MV's rowKey contract)

Niwat would happily take any of the three. Surfacing the `rowKey` as a helper is the most ergonomic — the consumer never has to know the underlying format.

### Net assessment

None of these block merging the spec — the pull-forward to v2 is a clean win for niwat. They're all "spec should resolve before #143 starts implementation" rather than "spec is wrong." The maintainer can address them inline or fold them into open implementation questions, depending on appetite.

For niwat planning purposes: assuming these are clarified, W3-A (PND.1 with MV + overlay), W3-B (DERIV-SSO-001 with junction + predicates + MV), W3-C (PP30 with same-collection partition MV) all become fully declarative in pre.14. That's a significant win over the original "MV + imperative shell" pre.14 plan — three rules collapsing from imperative shells to single declarative blocks.

vLannaAi:
**Resolved in `2d796ae`** — all five items from round 3 addressed.

### Merge-blocker (score 80+) — fixed

**Open question #3 stale cross-reference.** Updated the parenthetical to reflect that `declaredDeterministicPredicates` is in v2 (not v2.x), and to point forward at the actual sections rather than the now-removed "not supported" caveat. Both `predicateHash` and the canonical-JSON `ctx`-hash fold into `queryHash`.

### Design discussion (score 75) — addressed inline

**(1) `ctx` not folded into `queryHash`.** Picked niwat's preferred resolution (option 2): canonical-JSON hash of `ctx` folds into `queryHash` symmetric with `predicates.hash`. Updated JSDoc + new § "`ctx`-hash folding" subsection with the behavior table showing what changes hash-vs-not. Explicit warning about dynamic `ctx` (`new Date().toISOString()`) being an infinite-refresh hazard; `refresh: 'manual'` is the documented pattern for time-dependent MVs.

**(2) `QueryDependencyAnalyzer` overlay-awareness.** Added `OverlayResolver` to the algorithm sketch (analog of `joinResolver.resolveRef`) and a new `expand(name, deps, overlayResolver)` helper that resolves virtual collection names to `{base, overlay}` source pairs. New § "Overlay resolution" subsection. Shallow-expansion-only in v2 — transitive chains covered by cycle detection, not per-analysis recursion.

**(3) Overlay merge semantics for orphaned-override / overlay-only rows.** Operations table expanded from 5 rows to 12 rows covering all combinations of (base exists / overlay exists / predicate). Two key clarifications:
- Overlay-only + predicate true → return overlay (the orphaned-override case is intentional and supported — operators can preemptively write override rows before the base MV materializes)
- Overlay-only + predicate false → return `null` (stale/aborted override state)
- Delete-of-absent cases now explicit (idempotent on both sides)

New § "Orphaned-override pattern" with rationale + the consumer use case.

**(4) Overlay write id derivation.** New § "Virtual-collection writes derive `id` from the base MV's `rowKey`" — `OverlayedViewRegistry` resolves the base MV's `rowKey` at vault init and uses it for `vault.collection('pnd1').put(record)` calls. Consumers pass the row, not the id; the merge-by-id constraint is no longer something they have to know about. Escape hatch (`vault.collection('pnd1').overlay.rowKey(row)`) exposed for direct-overlay-collection writes.

### Spec growth

741 → 837 lines (+96). Most of the growth is the expanded operations table (12 rows × 4 columns) and the two new subsections (`ctx`-hash folding + orphaned-override pattern).

The spec should now be tight enough to gate the #143 implementation epic. Ready for round 4 of niwat-review or merge — your call.

vLannaAi:
### Code review

Found 2 issues — both internal contradictions in the same document, introduced when this commit pulled `withOverlayedView` into v2 + added the `put(record)` id-derivation API but didn't update older sections.

1. § Key invariants bullet still says `withOverlayedView` is "explicitly deferred to v2.5" — but commit `c857a6c` pulled it into v2 and the scope table at line 39 + § Composition at line 620 both confirm it ships in v2. The "deferred to v2.5" claim contradicts three other places in the same document.

https://github.com/vLannaAi/noy-db/blob/2d796aea4e8164964c8f4058317c644cd957235a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L126-L130

2. Integration test description still uses the old explicit-id `put(id, ...)` signature, but commit `2d796ae` introduced § "Virtual-collection writes derive id from the base MV's rowKey" (lines 695-711) that documents the API as `put(record)` (no id, derived internally). The test as written contradicts the new behavior documented just above it in the same commit.

https://github.com/vLannaAi/noy-db/blob/2d796aea4e8164964c8f4058317c644cd957235a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L771-L775

Spec is otherwise consistent and ready to land. Three smaller open questions surfaced in re-review (lazy-mode infinite-refresh, multi-overlay stacking pre-validation, explicit-id `put(id, record)` behavior) all scored below the merge-blocker threshold — happy to surface as a follow-up if useful before #143 starts.

vLannaAi:
**Follow-up — three round-4 design questions scored below merge-blocker threshold but worth resolving before #143 starts.**

### C. Infinite-refresh warning on dynamic `ctx` covers only `refresh: 'eager'`

The new § ctx-hash folding warns that dynamic `ctx` (e.g. `new Date()`) creates an infinite-refresh loop, and prescribes `refresh: 'manual'` as the escape. But the warning only addresses the eager case. `refresh: 'lazy'` hits the same problem via a different path:

1. Source write fires → stale bit set
2. Next read → executor recomputes `queryHash` with new `ctx` (e.g. current time) → mismatch → re-materialize
3. Next source write → stale bit set again → next read recomputes → mismatch → re-materialize
4. Same trap, just gated on read frequency instead of write frequency

Recommend extending the warning to say "dynamic `ctx` is incompatible with `eager` OR `lazy`; use `manual` and trigger refresh externally with each fresh `ctx` value." Otherwise an implementer who picks `lazy` thinking "lazy avoids the eager trap" walks into the same hole.

https://github.com/vLannaAi/noy-db/blob/2d796aea4e8164964c8f4058317c644cd957235a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L380-L470

### D. Pre-registration validation should reject `base` being itself a virtual overlay name

The spec says expansion in `analyzeDependencies` is "shallow" and that "No multi-overlay stacking" is a v2 non-goal. But pre-registration validation doesn't enforce the non-goal. A consumer accidentally writing:

```ts
withOverlayedView({ name: 'B', base: 'A', overlay: 'B-overrides', ... })
// where 'A' is itself a registered virtual overlay name
```

…passes pre-registration validation (because `A` is a known collection name in the registry), then produces incomplete dependency edges for any downstream MV that reads `B` — shallow expansion stops at `A`, never traces through to the concrete sources under `A`.

Cycle detection catches the case where the chain loops back, but doesn't catch the case where the chain just truncates downstream dep tracking. Recommend adding to § Pre-registration validation:

> `base` must reference a concrete collection — either a real source collection or an MV's output collection. A virtual overlay name is not permitted as `base` in v2 (multi-overlay stacking is a v3 feature; see § Scope deferred).

Throws something like `OverlayBaseIsVirtualError` at vault init. Closes the gap without committing to multi-overlay support.

https://github.com/vLannaAi/noy-db/blob/2d796aea4e8164964c8f4058317c644cd957235a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L405-L420

### E. `put(id, record)` with explicit id — behavior unspecified

The new § "Virtual-collection writes derive id from base MV's rowKey" documents the call as `vault.collection('pnd1').put(record)` — no id. But the operations table still shows `put(id, record)` as the canonical signature with a parenthetical "derives `id` if the consumer omits it." That implies the explicit-id form is supported. What does it do?

Three plausible behaviors that need disambiguation:

- (a) **Use verbatim.** `pnd1-overlay.put(someId, record)` writes the row with `id = someId`. If `someId !== rowKey(record)`, the merge silently creates an orphaned override.
- (b) **Validate against `rowKey(record)`.** Reject the write if they don't match.
- (c) **Reject explicit-id form entirely.** Force the `put(record)` shape.

Bulk-import use cases would lean toward (a) or (b). niwat would prefer (b) — pass-through when consistent, throw otherwise — because it makes typos surface immediately rather than producing silent orphan-override rows. Recommend one short paragraph picking a behavior.

https://github.com/vLannaAi/noy-db/blob/2d796aea4e8164964c8f4058317c644cd957235a/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L670-L715

### Net

None block merging. C is a one-line warning extension. D is one validation rule + one error type. E is one paragraph picking a behavior. Together they close the last latent ambiguities before the #143 implementation epic starts — making it less likely the implementer has to re-spec while writing code.

vLannaAi:
**Resolved in `a560edc`** — all five round-4 items addressed.

### Merge-blockers (Issues 1 + 2) — fixed

**Issue 1.** § Key invariants no longer claims `withOverlayedView` is "deferred to v2.5." Same stale claim found in § Materialized-collection shape (was outside Issue 1's call-out radius) — also updated to reflect v2 ships the primitive; arbitrary `mergePolicy` callbacks alone stay deferred to v3.

**Issue 2.** Integration test blurb for overlay write routing updated from `put(id, ...)` to `put(record)` with id-derivation explicit. Added a new test pinning the explicit-id validate-and-throw behavior (resolution for Issue E below).

### Follow-ups (Issues C + D + E) — also closed

**C — Infinite-refresh warning generalized.** The eager-only framing was misleading: lazy mode hits the same trap via the read-path (every read with a fresh `ctx` recomputes `queryHash` → constant mismatch → re-materialize per read). Updated to "Dynamic `ctx` is incompatible with `eager` AND `lazy`" with the two trap shapes documented side-by-side. `manual` remains the only correct refresh mode for time-dependent MVs.

**D — Pre-registration validation closes the virtual-base gap.** New § "Pre-registration validation (overlay)" under § Composition with operator-editable lifecycle, with three new error types:

- `OverlayBaseIsVirtualError(name, base)` — base must be concrete (real source or MV output)
- `OverlayCollectionUnavailableError(name, overlay)` — overlay must be a real, vault-known, non-MV-owned collection
- `OverlayNameCollisionError(name)` — virtual name must not collide with MV outputs or concrete sources

Closes the multi-overlay-stacking non-goal honestly at vault init rather than letting it ship as a latent dep-tracking truncation.

**E — `put(id, record)` validates, doesn't trust.** Picked niwat's preferred resolution: validate `id === rowKey(record)` and throw `OverlayIdMismatchError(id, expected)` on mismatch, pass-through on match. Preserves the standard `Collection<T>.put(id, record)` signature for generic code paths; surfaces typos immediately rather than producing silent orphans. Direct writes to the underlying overlay collection bypass the validation (the constraint is on the virtual layer only — bulk-import paths retain full id control). New § "Explicit-id `put(id, record)` — validate, don't trust" under the existing virtual-collection-writes section.

### Spec growth

837 → 864 lines (+27). Last cleanup commit on this PR (assuming no more contradictions surface).

Spec should now be ready for merge → unblocks the #143 implementation epic.

vLannaAi:
### Code review

Found 1 issue (score ≥ 80):

1. Four new overlay error types are introduced in prose (`OverlayIdMismatchError`, `OverlayBaseIsVirtualError`, `OverlayCollectionUnavailableError`, `OverlayNameCollisionError`) but none are added to the § Error handling table. That table is the spec's authoritative error reference and currently lists only MV-phase errors. An implementer using the table as a checklist for `packages/hub/src/errors.ts` would miss all four.

https://github.com/vLannaAi/noy-db/blob/a560edc83153a164d3b544a1b0cc99a7e0e9acd2/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L765-L780

One additional concern (score 75): `OverlayIdMismatchError` constructor signature is inconsistent across two locations — line 681 shows `OverlayIdMismatchError(id, expected)` and line 726 shows `OverlayIdMismatchError(id, rowKey(record))`. Below the merge-blocker threshold but worth picking a canonical signature before #143 starts.

https://github.com/vLannaAi/noy-db/blob/a560edc83153a164d3b544a1b0cc99a7e0e9acd2/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L679-L683

https://github.com/vLannaAi/noy-db/blob/a560edc83153a164d3b544a1b0cc99a7e0e9acd2/docs/superpowers/specs/2026-05-20-dim14-mv-v2-design.md#L724-L728

Otherwise spec is consistent. All round-4 follow-ups (A through E) addressed as requested. Spec is ready to land after the error-table addition.

vLannaAi:
**Resolved in `992284c`** — both round-5 items closed.

### Merge-blocker — overlay errors added to § Error handling table

Four new rows appended for the round-4-introduced overlay errors:

- `OverlayBaseIsVirtualError(name, base)` — multi-overlay-stacking attempt
- `OverlayCollectionUnavailableError(name, overlay)` — overlay collection unknown or MV-owned
- `OverlayNameCollisionError(name)` — virtual name collides with MV output or concrete source
- `OverlayIdMismatchError(actual, expected)` — virtual `put(id, record)` with `id !== rowKey(record)`

Implementers using the table as a checklist for `packages/hub/src/errors.ts` will now see them.

### Sub-blocker — `OverlayIdMismatchError` signature normalized

Picked `(actual, expected)` as the canonical signature — names match the validation semantics. Updated three reference sites:
- Operations-table row 681
- Prose at line 726 (now: "`actual` is the consumer-supplied `id`, `expected` is `rowKey(record)`")
- New § Error handling row (line 782)

All three now agree.

### Spec growth

864 → 869 lines (+5). Smallest cleanup commit in the round series.

Spec should be ready for merge.

vLannaAi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

PR #158, opened 2026-05-20
feat(hub): MV correctness — partition, onEmpty, ceiling, strict, aggregate (#152)

vLannaAi:
## Summary

Stacks on PR #157 (lazy + manual). Closes #152 — sub-task 3 of the #143 implementation epic. Spec sequencing steps 9–12 + the aggregate-query shape (deferred from #150's foundation).

## What's new

- **Cost ceiling**: \`MaterializedViewTooLargeError\` fires before writes when row count exceeds \`maxRows\` (default 100k, overridable)
- **Tombstoning**: \`onEmpty: 'delete'\` (default) tombstones disappeared rows via \`_internalDelete\` (bypasses user \`onDelete\` per PR #148's composition fix); \`onEmpty: 'keep'\` opts out
- **Same-collection partition**: \`output.partition: { field, value }\` discriminator + cycle detector resolution via provability check (supports \`==\` / \`!=\` / \`in\`); rejects naïve same-collection MVs as \`MaterializedViewCycleError\`
- **Strict-mode rollback**: \`strict: true\` re-throws on row failure → TxContext rollback via \`revertExecuted\` (#133)
- **Aggregate query shape**: executor branches on \`.toArray()\` / \`.run()\` terminals; aggregate MVs require explicit \`sources\` since the analyzer can't walk through \`groupBy().aggregate()\` back to the source

## Bundle

Essentially flat (floor -0.1% vs the new baseline #151 established). #152 extends existing primitives — no new heavy files.

## Stacked PR base

Targets \`feat/mv-lazy-manual-151\` so the diff is the #152-only delta. GitHub auto-rebases onto main once #156 and #157 merge.

## Test plan

- [x] \`pnpm --filter @noy-db/hub typecheck\`
- [x] \`pnpm --filter @noy-db/hub lint\` — 0 errors
- [x] \`pnpm --filter @noy-db/hub test\` — 1546 / 1546 (was 1534 + 12 new)
- [x] \`pnpm --filter @noy-db/hub bundle-check\` — all four scenarios green

Closes #152.

PR #159, opened 2026-05-20
feat(hub): MV declaredDeterministicPredicates + .wherePredicate (#153)

### LINKED ISSUE #153: feat(hub): declaredDeterministicPredicates + .wherePredicate(name, ctx?) (#143 subtask)
Sub-issue of #143. Spec sequencing step 13.

## Scope

- New `predicates?: { [name]: { hash, fn } }` field on `MaterializedViewStrategy`
- New `QueryBuilder.wherePredicate(name, ctx?)` terminal
- `predicateHash` folds into `queryHash` — bumping `hash` forces refresh on next visit
- `ctx` argument: canonical-JSON serialization, SHA-256 folded into `queryHash` symmetric with `predicates.hash`
- Refresh fires when EITHER the predicate's `hash` bumps OR the canonical `ctx` changes

## Acceptance

- `withMaterializedView({ predicates: { isOverdue: { hash, fn } }, query: () => invoices.query().wherePredicate('isOverdue', { asOf }) ... })` registers + materializes
- `queryHash` changes when `predicates.isOverdue.hash` is bumped → refresh fires on next read
- `queryHash` changes when literal `ctx` value changes (e.g. `{ asOf: 'A' }` → `{ asOf: 'B' }`) → refresh fires on next read
- `queryHash` unchanged when `ctx` keys are reordered (canonical serialization)
- Predicate fn receives `(row, ctx)`; gets the full cross-joined row shape from any upstream `.join()` clause
- Composition with `.groupBy()` / `.aggregate()` / standard `.where()` works

## Open implementation question

The spec's § "Function-based source-row predicates" explicitly notes that *dynamic* `ctx` (e.g. `{ asOf: new Date().toISOString() }`) creates an infinite-refresh loop on `eager` AND `lazy`. The implementation should NOT try to detect this — the consumer is responsible. Document the trap in the JSDoc.

## Depends on

The MV foundation sub-issue (steps 1-6); independent of lifecycle/correctness sub-issues.

vLannaAi:
Stacks on PR #158. Closes #153 — sub-task 4 of #143. Spec sequencing step 13.

## What's new

- `MaterializedViewStrategy.predicates: { [name]: { hash, fn } }` field
- `Query.wherePredicate(name, ctx?)` builder terminal
- New `WherePredicateClause` in the query DSL — carries resolved fn + hashes at construction time
- queryHash folds both `predicate.hash` and a canonical-JSON `ctxHash` symmetrically

## How it composes

The MV registry wraps the `db` passed to `spec.query(db)` so `.collection().query()` returns a Query<T> with predicates attached. Bare Queries (outside an MV) throw on `.wherePredicate()` with an actionable error pointing to the strategy field.

## Verification

- 1551 hub tests pass (was 1546 + 5 new)
- typecheck clean · lint clean
- Bundle essentially flat — predicate machinery is just two new clause-case branches and a wrapper proxy

Closes #153.

### DIFF
--- a/packages/hub/__tests__/materialized-views/predicates.test.ts
+++ b/packages/hub/__tests__/materialized-views/predicates.test.ts
@@ -1,0 +1,195 @@
+import { describe, it, expect } from 'vitest'
+import { createNoydb, withMaterializedView } from '../../src/index.js'
+import type { NoydbStore, EncryptedEnvelope } from '../../src/types.js'
+
+function memory(): NoydbStore {
+  const data = new Map<string, EncryptedEnvelope>()
+  const k = (v: string, c: string, i: string) => `${v}/${c}/${i}`
+  return {
+    capabilities: { casAtomic: true, auth: { kind: 'none' } },
+    async get(v, c, i) { return data.get(k(v, c, i)) ?? null },
+    async put(v, c, i, env) { data.set(k(v, c, i), env) },
+    async delete(v, c, i) { data.delete(k(v, c, i)) },
+    async list(v, c) {
+      const prefix = `${v}/${c}/`
+      return [...data.keys()].filter(key => key.startsWith(prefix)).map(key => key.slice(prefix.length))
+    },
+    async loadAll(v) {
+      const out: Record<string, Record<string, EncryptedEnvelope>> = {}
+      for (const [key, env] of data) {
+        const [vname, cname, id] = key.split('/')
+        if (vname === v) {
+          out[cname] = out[cname] ?? {}
+          out[cname][id] = env
+        }
+      }
+      return out
+    },
+    async saveAll(v, payload) {
+      for (const c of Object.keys(payload)) {
+        for (const i of Object.keys(payload[c])) {
+          data.set(k(v, c, i), payload[c][i])
+        }
+      }
+    },
+  }
+}
+
+interface Invoice extends Record<string, unknown> {
+  id: string
+  status: 'open' | 'paid'
+  amount: number
+  dueDate: string
+}
+
+describe('MV declaredDeterministicPredicates (#153)', () => {
+  it('wherePredicate filters rows via the registered fn', async () => {
+    const mv = withMaterializedView<Invoice>({
+      name: 'overdue',
+      predicates: {
+        isOverdue: {
+          hash: 'is-overdue-v1',
+          fn: (inv: Invoice, ctx?: unknown) => {
+            const { asOf } = ctx as { asOf: string }
+            return inv.status === 'open' && inv.dueDate < asOf
+          },
+        },
+      },
+      query: (db) => db.collection<Invoice>('invoices').query().wherePredicate('isOverdue', { asOf: '2026-05-20' }),
+      rowKey: (r) => r.id,
+      refresh: 'eager',
+    })
+    const db = await createNoydb({
+      store: memory(),
+      user: 'alice',
+      secret: 'mv-predicates-basic-passphrase-2026',
+      materializedViewStrategies: [mv],
+    })
+    const vault = await db.openVault('demo')
+    await vault.collection<Invoice>('invoices').put('a', { id: 'a', status: 'open', amount: 100, dueDate: '2026-05-01' })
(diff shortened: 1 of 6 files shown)


vLannaAi:
### Code review

Found 2 issues:

1. Every chain method that returns a new `Query<T>` drops `this.predicates`. Only `_withPredicates()` and `.wherePredicate()` itself thread the predicates map through to the new Query — `where()`, `or()`, `and()`, `filter()`, `orderBy()`, `limit()`, `offset()`, and `join()` all reconstruct `new Query<T>(source, plan, joinContext, aggregateStrategy)` with only four arguments, so the 5th-slot `predicates` becomes `undefined`. Any composition like `q.where('id', '==', x).wherePredicate('isPaid')` throws `no predicates registered on this Query`. The tests pass because they all call `.wherePredicate()` immediately after `.query()`.

`wherePredicate` (preserves predicates — 5 args):

https://github.com/vLannaAi/noy-db/blob/fa0be50a032e70904a62edac08a2ffd54c368677/packages/hub/src/query/builder.ts#L177-L210

`where` (drops predicates — 4 args):

https://github.com/vLannaAi/noy-db/blob/fa0be50a032e70904a62edac08a2ffd54c368677/packages/hub/src/query/builder.ts#L210-L220

Same issue at: `or` L226, `and` L247, `filter` L265, `orderBy` L279, `limit` L289, `offset` L299, plus the `join` legs further down.

2. `serializeClause` does not handle `'wherePredicate'`. The 'filter' and 'group' cases strip/recurse, but a `WherePredicateClause` falls through to `return clause`, exposing the live `fn` reference. The `QueryPlan` doc comment claims plans are JSON-serializable "as long as no FilterClause is present" — `WherePredicateClause` is now a second non-serializable shape that the invariant comment doesn't mention. `JSON.stringify(serializePlan(plan))` will silently drop the function and produce identical serializations for distinct predicates.

https://github.com/vLannaAi/noy-db/blob/fa0be50a032e70904a62edac08a2ffd54c368677/packages/hub/src/query/builder.ts#L941-L956

Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

vLannaAi:
**Resolved in `b171cce`** (rebased into the stack).

### #1 — Chain methods now thread `this.predicates`
`.where()`, `.or()`, `.and()`, `.filter()`, `.orderBy()`, `.limit()`, `.offset()`, `.join()` — all 9 `new Query<T>(...)` construction sites in builder.ts now pass `this.predicates` as the 5th argument. Tests had a blind spot — all called `.wherePredicate()` immediately after `.query()`, never through a chain step. New regression test composes `where → wherePredicate → orderBy → limit` to prevent recurrence.

### #2 — `serializeClause` handles `wherePredicate`
New case added alongside the existing `filter` case. Strips `fn` (placeholder `'[function]'`), keeps `name`/`ctx`/`predicateHash`/`ctxHash` — distinct predicates now serialize distinctly via their identity-carrying fields. The QueryPlan invariant (non-serializable when FilterClause is present) extends naturally to WherePredicateClause.

vLannaAi:
### Code review

No issues found. Both prior findings are addressed at https://github.com/vLannaAi/noy-db/commit/b171cce5d8f3d170de583d622d32088280d62bd2:

- `where`, `or`, `and`, `filter`, `orderBy`, `limit`, `offset`, and `join` now thread `this.predicates` into the new Query (https://github.com/vLannaAi/noy-db/blob/b171cce5d8f3d170de583d622d32088280d62bd2/packages/hub/src/query/builder.ts#L210-L221)
- `serializeClause` now strips `WherePredicateClause.fn` to `'[function]'` while keeping `name`/`predicateHash`/`ctxHash` (https://github.com/vLannaAi/noy-db/blob/b171cce5d8f3d170de583d622d32088280d62bd2/packages/hub/src/query/builder.ts#L951-L972)

Generated with [Claude Code](https://claude.ai/code)

<sub>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #167, opened 2026-05-21
feat(hub): pre.15 — multi-key groupBy (#166) + UNION MV (#165) + guards variance (#131)

### LINKED ISSUE #131: refactor(hub): refine GuardStrategyHandle type variance to remove public-API any
## Context

The pre.11 guards feature shipped with two `eslint-disable-next-line @typescript-eslint/no-explicit-any` annotations on public-API option fields:

- `packages/hub/src/vault.ts:334` — Vault constructor option `guardStrategies?: ReadonlyArray<GuardStrategyHandle<any>>`
- `packages/hub/src/types.ts:1748` — `NoydbOptions.guardStrategies?: ReadonlyArray<GuardStrategyHandle<any>>`

## Why `any` was used

Narrowing to `unknown` or a concrete `Record<string, unknown>` bound breaks consumer-side type inference. Showcase 79 passes `GuardStrategyHandle<Invoice>` and `GuardStrategyHandle<Disbursement>` to `guardStrategies: [invoiceGuard, disbursementGuard]`. The handle's `T` parameter appears in callback positions (`check(incoming: T, ctx)`, `invariant(changes: ReadonlyArray<GuardChange<T>>, ctx)`) so `GuardStrategyHandle<T>` is *invariant* in `T` — `Handle<Invoice>` is not assignable to `Handle<unknown>`.

## Possible refactors

1. **Phantom-type position:** Move `T` out of callback positions inside the handle (use a discriminant tag plus `unknown` payloads internally; keep the typed shape only in the public `withGuard` factory).
2. **Existential boundary:** Define a sealed `GuardStrategyHandleAny` interface as the array element type — TS doesn't have first-class existentials, but a structurally narrow shape can fake it.
3. **Indexed-access workaround:** Type the option as `ReadonlyArray<GuardStrategyHandle<infer-any-record>>` via a helper conditional type that erases the parameter.

All three preserve the consumer's ability to pass strongly-typed handles AND eliminate the `any` from public APIs.

## Acceptance

- Two `eslint-disable-next-line` annotations removed
- Showcase 79 type-checks unchanged
- All guards tests still pass
- No new `any` introduced elsewhere

## References

- Branch `feat/guards` (commit `bba866a` HEAD), specifically `vault.ts:334` and `types.ts:1748`
- Surfaced by T15 final sweep in the pre.11 milestone execution

### LINKED ISSUE #165: feat(hub): UNION MV — withMaterializedView reading from multiple sibling collections
## Background

Pre.14 \`withMaterializedView\` accepts a single \`query()\` that walks one source collection (or joins from one source). For aggregates that need to **sum across two sibling collections** — same row shape, different storage namespace — there is no first-class way today.

## Real consumer need (niwat)

Niwat issues two legally-distinct payment-related documents:

- \`taxReceipts\` (combined ใบกำกับภาษี/ใบเสร็จรับเงิน, §86) — positive amounts on issuance
- \`creditNotes\` (ใบลดหนี้, §86/10) — corrections referencing an original tax receipt

Both contribute to **Niwat's monthly output VAT obligation (PP.30)** via the same field:

\`\`\`ts
monthlyOutputVat(period) =
    Σ taxReceipt.paidServicesVat   WHERE issuedAt in period
  + Σ creditNote.paidServicesVat    WHERE issuedAt in period   // negatives reduce the period
\`\`\`

The UNION is the natural query. Today the workaround is **two MVs the consumer sums** at read time:

\`\`\`ts
const taxReceiptVatByMonth = withMaterializedView<{ period: string; vat: number }>({
  name: 'taxReceiptVatByMonth',
  query: db => db.collection('taxReceipts').query().groupBy('period').aggregate({ vat: sum('paidServicesVat') }),
  // ...
})
const creditNoteVatByMonth = withMaterializedView({ /* same shape on creditNotes */ })

// Consumer:
const total = (taxReceiptVatByMonth.get(period)?.vat ?? 0) + (creditNoteVatByMonth.get(period)?.vat ?? 0)
\`\`\`

This works but pushes the union into product code, defeats the "no imperative transforms outside primitives" rule niwat is trying to enforce (§ VIEW-DECL-001), and means every consumer site has to remember to add both arms.

## Proposed API

Three shapes were considered. **Option 1 is the niwat-recommended path** — see Consumer Recommendation section below.

### Option 1 — explicit \`unionSources\` field ⭐ RECOMMENDED

\`\`\`ts
withMaterializedView<MonthlyVatRow>({
  name: 'monthlyOutputVat',
  unionSources: [
    { collection: 'taxReceipts', map: row => ({ period: row.issuedAt.slice(0, 7), vat: row.paidServicesVat }) },
    { collection: 'creditNotes', map: row => ({ period: row.issuedAt.slice(0, 7), vat: -row.vatDifference }) },
  ],
  groupBy: ['period'],
  aggregate: { vat: sum('vat'), count: count() },
  rowKey: row => row.period,
  refresh: 'eager',
})
\`\`\`

The executor reads each source, runs the per-source \`map\` to unify the row shape, concatenates, then runs groupBy + aggregate. Source-write hooks fire on EITHER source.

**Constraint: \`unionSources\` and \`query\` are mutually exclusive** — an MV either does a single-source query (existing path) OR a union (new path), never both. When \`unionSources\` is set, the aggregation spec lives at the MV-strategy level (\`groupBy\` / \`aggregate\` as siblings of \`unionSources\`).

### Option 2 — \`.union(otherQuery)\` chain operator

\`\`\`ts
query: db =>
  db.collection('taxReceipts').query()
    .map(r => ({ period: r.issuedAt.slice(0, 7), vat: r.paidServicesVat }))
    .union(
      db.collection('creditNotes').query()
        .map(r => ({ period: r.issuedAt.slice(0, 7), vat: -r.vatDifference })),
    )
    .groupBy('period')
    .aggregate({ vat: sum('vat') })
\`\`\`

Composable with existing query builder. Dep analyzer walks both arms. More flexibility but more places for dep-analysis bugs — niwat-review pre-merge has historically caught composition bugs in chain operators (e.g., \`predicates\` map dropping through \`.where()\` in pre.14 #153).

### Option 3 — \`db.unionOf\` accessor

Reuse the existing \`sources?: string[]\` array; the query callback receives a special \`db.unionOf(['taxReceipts', 'creditNotes'])\` accessor that emits a unified stream the consumer can groupBy.

Lightest API change. Conflates two concepts: \`sources\`-as-deps (cache invalidation) vs \`sources\`-as-union-arms.

## Consumer recommendation (niwat)

**Option 1.** Three reasons:

1. **Declarative shape matches existing MV definitions.** \`withMaterializedView\` is already a spec object (\`name\`, \`query\`, \`rowKey\`, \`sources\`, \`refresh\`). \`unionSources\` slots in symmetrically. Option 2's chained \`.union()\` would be the only thing in the strategy file that isn't a structured spec field.

2. **Per-source \`map\` is exactly where it should be.** Niwat's real use case unions \`taxReceipts\` (with \`paidServicesVat\` field) and \`creditNotes\` (with \`vatDifference\` field) — **different row shapes**. Option 1 forces the consumer to write the unification explicitly inside the strategy. Option 2/3 push that into either a chained \`.map()\` or a callback closure; either way the unification step is less reviewable.

3. **Lowest upstream landing risk.** Option 2 multiplies the AST-walking surface (dep analyzer + source-write hook registry both need to traverse two query arms). Option 1's registry just reads \`strategy.unionSources.map(s => s.collection)\` directly. Smaller test matrix, faster ship.

**Trade-off accepted:** niwat loses the ability to express "filter both arms, then union, then aggregate" without prefilters baked into each \`map\` callback. All known niwat use cases are simple \`union → groupBy → aggregate\` — no need for the extra composability.

## Constraints / questions

- **Schema compatibility.** The two collections may have different schemas. The \`map\` step is mandatory; the executor can't assume the row shapes overlap.
- **Dependency analyzer.** Both sources must show up in \`sourceVersions\` of the materialized rows (so a write to either re-fires the MV refresh).
- **Source-write hook.** \`Collection.put\` on either source must trigger the same refresh path. The registry needs reverse-index from \`collection → MV\` to look up MVs that union the touched source.
- **Strict / onEmpty / maxRows.** All existing fields apply unchanged.
- **No multi-key groupBy** — orthogonal to this issue. Tracked separately as #166 (the two compose naturally — niwat needs both for client-side tax-document aggregates).

## Niwat tracking

This blocks the canonical monthly-VAT MV in niwat's refactor (filed as niwat issue #79). Without UNION MV, every consumer that needs monthly output VAT either:
1. Reads both per-collection MVs and sums client-side (current workaround plan)
2. Continues with the imperative \`.filter().reduce()\` reducer in stores/composables
3. Triggers a derivation that writes a \`vatContribution\` sibling row from BOTH sources (expensive — N derivations, one per source row)

## Urgency

Niwat is mid-refactor (vault-strategies module landed in niwat commit \`9fc7ee4\` for niwat issue #81 + sibling #82). The taxReceipts ↔ creditNotes split is happening now; a UNION MV would slot in cleanly during that work. Without it, niwat ships with a two-MV workaround that we'd have to migrate later.

## Acceptance

- \`unionSources\` field accepted on \`withMaterializedView\` strategy spec.
- \`unionSources\` and \`query\` are mutually exclusive (validation at registration time throws a clear error if both are set).
- When \`unionSources\` is set, sibling fields \`groupBy: string[]\` and \`aggregate: {...}\` declare the post-union aggregation.
- Per-source \`map\` produces row shape consumed by \`groupBy\` + \`aggregate\`.
- Source-write hook fires on Collection.put for EITHER source listed in \`unionSources\`.
- \`_materializedFrom.sourceVersions\` records the version of each contributing source.
- Hub test exercises a two-source UNION + groupBy + aggregate, with source-write hooks firing on both arms.
- A new showcase under \`showcases/src/\` documents the pattern.
- \`features.yaml\` entry added.
- Subsystem doc updated.

## Related

- #166 — multi-key groupBy (composes naturally with this; niwat's client tax-document MV needs both: \`union(...).groupBy('clientId', 'period')\`)
- vLannaAi/niwat#79 — niwat monthly receipt aggregates MV (the primary consumer of this feature)
- vLannaAi/niwat#81, vLannaAi/niwat#82 — the taxReceipts ↔ creditNotes split that makes UNION necessary

### LINKED ISSUE #166: feat(hub): multi-key groupBy — withMaterializedView / Query.groupBy(...fields)
## Background

Pre.14 \`Query<T>.groupBy<F extends string>(field: F)\` takes a **single** field. The returned \`GroupedQuery<T, F>\` and \`GroupedAggregation<R>\` carry a single \`field: F\` internally. No first-class way to express \`SQL GROUP BY a, b\`.

The downstream effect bites whenever an MV needs **one row per composite key** — the canonical "per-(tenant, period)" roll-up shape.

## Real consumer pain (niwat)

Two niwat MVs are currently blocked on this:

### niwat issue #77 — auto-disbursement (PND.1 + SSO)

\`\`\`ts
// Want:
withMaterializedView<Pnd1Row>({
  name: 'pnd1-auto',
  query: db => db.collection<Compensation>('compensations')
    .query()
    .groupBy('clientId', 'period')        // ← single-key today
    .aggregate({ total: sum('taxAmount'), count: count() }),
  rowKey: row => \`\${row.clientId}|\${row.period}\`,
  sources: ['compensations'],
  refresh: 'eager',
})
\`\`\`

Output collection needs one row per \`(clientId, period)\`. Today's workaround:

1. Add a synthetic \`clientPeriodKey: string\` field to the \`Compensation\` schema (computed as \`\${clientId}|\${period}\` at every write site)
2. \`.groupBy('clientPeriodKey')\` — single-key works
3. Caller decodes \`clientPeriodKey\` back to \`{clientId, period}\` on read

Cost: every write to Compensation must remember to populate \`clientPeriodKey\`; the field is denormalised data that any code path can desync. The MV's "one fact in one place" guarantee gets pierced.

### niwat issue #84 — client tax-document aggregates

Same shape, different source. The new \`TaxDocument\` collection needs MVs aggregating to \`vatSales\` / \`vatPurchase\` / \`pnd3\` / \`pnd53\` totals per \`(clientId, period)\` keyed by direction:

\`\`\`ts
// Want:
.query()
.groupBy('clientId', 'period', 'direction')
.aggregate({ vatTotal: sum('vatAmount'), whtTotal: sum('whtAmount'), count: count() })
\`\`\`

The three-key composite (\`clientId\`, \`period\`, \`direction\`) makes the synthetic-key workaround uglier — \`\${clientId}|\${period}|\${direction}\` strings ripple through every consumer.

## Proposed API

Variadic \`groupBy(...fields)\`. Backward-compatible with the single-arg form.

\`\`\`ts
// Today (still works):
.groupBy('clientId').aggregate({ total: sum('amount') })
//   → row shape: { clientId: string, total: number }

// New:
.groupBy('clientId', 'period').aggregate({ total: sum('amount') })
//   → row shape: { clientId: string, period: string, total: number }

.groupBy('clientId', 'period', 'direction').aggregate({ vatTotal: sum('vatAmount') })
//   → row shape: { clientId: string, period: string, direction: string, vatTotal: number }
\`\`\`

### Type signature sketch

\`\`\`ts
// Single field — keeps current narrow-typed return for back-compat
groupBy<F extends string>(field: F): GroupedQuery<T, F>

// Multiple fields — variadic
groupBy<F extends readonly [string, ...string[]]>(
  ...fields: F
): GroupedQueryN<T, F>

// GroupedQueryN.aggregate emits a row containing all grouped fields + the aggregate output
\`\`\`

The query-hash algorithm canonicalises field order (sorted) so \`.groupBy('a', 'b')\` and \`.groupBy('b', 'a')\` produce identical hashes / output rows.

### Live mode

\`.groupBy(...).aggregate().live()\` already runs the full aggregation per source-write under \`live()\`. Multi-key changes nothing in that loop — the cardinality is just (distinct-tuples) instead of (distinct-singles).

### Cardinality warnings

The existing single-key threshold (\`groupby.ts\` line ~93) warns at high group cardinality. Multi-key inherits the same threshold; the warning message lists all grouped fields.

## Constraints

- **Composite \`rowKey\`** — the MV consumer decides the row-id format. The hub does NOT impose a canonical encoding. The strategy's \`rowKey: row => ...\` callback receives the full row (with all grouped fields populated), so consumers can format however they like (\`\${a}|\${b}\`, JSON canonicalisation, hash, etc.).
- **Index hints** — same as single-key: groupBy can't be index-accelerated (the warning in the existing code applies). Multi-key just multiplies the candidate count by combinations; no new algorithmic complexity.
- **Determinism** — same per-MV \`queryHash\` rules apply. Field-order canonicalisation ensures stable hashes.

## What blocks if this doesn't land

Niwat ships **#77** and **#84** with synthetic-key workarounds:
- Compensation gains a \`clientPeriodKey\` field
- TaxDocument gains a \`clientPeriodDirectionKey\` field
- Every write site (CRUD actions, mock seed, scan-upload flow) must populate these
- Any new code path that forgets the synthetic key silently produces wrong MV output

A multi-key groupBy lets niwat write \`groupBy('clientId', 'period')\` cleanly and stop maintaining the synthetic fields.

## Adjacent feature — same release?

This composes naturally with **noy-db #165 (UNION MV)**. Niwat's monthly-VAT MV wants both: \`union(taxReceipts, creditNotes).groupBy('clientId', 'period')\`. If the two land in the same release the niwat refactor can pull them together; otherwise the consumer-side workarounds stack (synthetic key AND two-MV summing).

## Acceptance

- \`Query<T>.groupBy(...fields)\` accepts 1..N fields. Single-field call site keeps its current narrowed type.
- \`.aggregate({...})\` emits rows containing ALL grouped fields plus the aggregate result.
- \`withMaterializedView\` consumer uses the variadic call inside \`query()\`; row shape's grouped fields appear correctly stamped.
- Cardinality warning fires when distinct tuples exceed the existing threshold.
- Showcase under \`showcases/src/\` exercises a 2-key and a 3-key groupBy with eager refresh, with assertions on the output row shape.
- \`features.yaml\` and \`docs/subsystems/derivations.md\` updated to document multi-key.

## References

- Existing single-key impl: \`packages/hub/src/query/builder.ts\` (groupBy method) + \`packages/hub/src/aggregate/groupby.ts\` (GroupedQuery / GroupedAggregation)
- Niwat issues blocked: vLannaAi/niwat#77, vLannaAi/niwat#84
- Adjacent request: #165 (UNION MV)

### LINKED ISSUE #169: fix(hub): UNION MV silently ignores `aggregate` when `groupBy` is absent
Surfaced in PR #167 code review. Not blocking pre.15 release but a silent failure mode worth closing.

## Problem

In `packages/hub/src/materialized-views/executor.ts`, `materializeUnionResult` early-returns the raw concatenated rows when `spec.groupBy` is falsy — even if `spec.aggregate` is set:

```ts
// executor.ts:112
if (!spec.groupBy) return unified
// spec.aggregate is never consulted
```

The validation in `with-materialized-view.ts` only rejects an empty-array `groupBy` (`groupBy: []`); it has no guard for the `{ aggregate, /* no groupBy */ }` combination. Registration succeeds, the executor silently skips the reducer, and the MV is populated with un-aggregated rows.

No error, no warning, wrong shape. Consumer sees a Collection of raw concatenated rows when they expected one row per arm with the reducer applied.

## Repro

```ts
withMaterializedView({
  name: 'totals_per_company',
  unionSources: [
    { collection: 'taxReceipts', map: r => ({ companyId: r.companyId, amount: r.amount }) },
    { collection: 'creditNotes', map: r => ({ companyId: r.companyId, amount: -r.amount }) },
  ],
  // groupBy intentionally omitted
  aggregate: { total: sum('amount') },
})
```

Expected: registration throws, OR aggregate runs as a single non-grouped reduction.
Actual: registration succeeds, MV output collection contains raw un-aggregated rows.

## Suggested fix

Add at registration in `with-materialized-view.ts` (matching the pattern from `5beaaa6` which rejects empty-array `groupBy`):

```ts
if (spec.unionSources && spec.aggregate && !spec.groupBy) {
  throw new MaterializedViewConfigError(
    'UNION MV with `aggregate` requires `groupBy` — ' +
    'use groupBy to declare the bucketing keys, or remove aggregate for a pure dedup MV'
  )
}
```

Alternative: run the aggregate as a single non-grouped reduction (one output row) when `aggregate` is set without `groupBy`. The error path is safer — less magic, surfaces the misconfiguration immediately.

## Affected files

- `packages/hub/src/materialized-views/with-materialized-view.ts` — add the guard
- `packages/hub/__tests__/materialized-views/union-validation.test.ts` — pin the rejection

### LINKED ISSUE #170: fix(hub): `predicates` silently no-ops on UNION MVs (hash + execution gap)
Surfaced in PR #167 code review. Not blocking pre.15 release but a silent failure mode that breaks the contract established by #153/#159.

## Problem

`spec.predicates` is silently ignored for UNION-mode MVs, in two places.

**1. Hash gap** — `summarizeUnionPlan` (`packages/hub/src/materialized-views/dependency-analyzer.ts`) does not fold `spec.predicates[*].hash` into the hash output, unlike `summarizeQueryPlan` which was specifically designed for this (see commit `d23372a` — "MV declaredDeterministicPredicates + .wherePredicate"). A consumer who declares `predicates` on a UNION strategy and bumps a predicate's `hash` sees no change in `queryHash`, so stale rows are not detected.

**2. Execution gap** — `materializeUnionResult` in `executor.ts` calls `coll.query().toArray()` for each arm. It never invokes `.wherePredicate(name)`. The `predicate-wrapped` `db` is passed in but its wrap is inert because the query chain never opts in. Predicates declared on a UNION strategy simply do not execute.

Together: a consumer who tries `predicates` on a UNION strategy gets no error, the predicate never runs, and the hash never propagates. Both invariants from #153/#159 are violated for the new UNION path.

## Suggested fix

The pattern from `5beaaa6` (and now the arm-order fix in `6be47a29`) is clear: when a config combination is unsupported in UNION mode, reject it at registration with `MaterializedViewConfigError`. Predicates are non-trivial to support on UNION (each arm would need its own predicate set or a unified semantic) — easier to gate them out cleanly until there's a need.

In `with-materialized-view.ts`:

```ts
if (spec.unionSources && spec.predicates) {
  throw new MaterializedViewConfigError(
    'predicates is not supported on UNION strategies yet — ' +
    'UNION mode does not use a Query<T> chain, so .wherePredicate() cannot fire. ' +
    'Use query() form, or open an issue if you need this.'
  )
}
```

If predicates on UNION MVs becomes a needed feature later, the proper fix is (a) fold `spec.predicates` into `summarizeUnionPlan` and (b) define per-arm `.wherePredicate` invocation semantics.

## Affected files

- `packages/hub/src/materialized-views/with-materialized-view.ts` — add the guard
- `packages/hub/__tests__/materialized-views/union-validation.test.ts` — pin the rejection

## Related

- #153 — MV declaredDeterministicPredicates + .wherePredicate
- #159 — predicates chain method propagation
- PR #167 — pre.15 UNION MV (introduced the gap)

vLannaAi:
## Summary

Pre.15 cycle landing three issues on one branch (#166, #165, #131). #165 + #166 together unblock the niwat refactor's monthly-VAT MV; #131 is a pre.14 type-only cleanup. Two niwat review follow-ups (#169, #170) folded in as small guards.

- **Spec:** `docs/superpowers/specs/2026-05-22-dim14-mv-multikey-and-union-design.md`
- **Plan:** `docs/superpowers/plans/2026-05-22-dim14-mv-multikey-and-union.md`

### #166 — Variadic Query.groupBy(...fields)
- New variadic overload on `Query<T>.groupBy(...)` with back-compat single-arg
- `canonicalGroupKey` helper for stable dedup (sorted field names)
- Result rows carry every grouped field in DECLARATION order + reducer outputs
- Cardinality warning lists every grouped field name; threshold unchanged
- MV strategies can use multi-key groupBy inside `query()` callback unchanged
- New showcase 85, features.yaml entries under `aggregate` + `materialized-views`, sections in `aggregate.md` + `derivations.md`

### #165 — UNION MV via unionSources
- New `unionSources: [{ collection, map }, ...]` field on `withMaterializedView` (Option 1 from issue)
- Per-source `map` is the schema-unification boundary — sibling collections with different schemas project to one MV row shape
- Declarative `groupBy: string | string[]` + `aggregate` fields run on the concatenated stream
- Mutually exclusive with `query`; ≥2 arms required; distinct collection names required; empty `groupBy: []` rejected
- New `MaterializedViewConfigError` for malformed strategies
- Source-write hook fires on every arm via the existing `dependencies` reverse-index (no `Collection.put` changes)
- `summarizeUnionPlan` hashes arm collection names in **declaration order** (semantically meaningful for the dedup-only path); `groupBy` fields + `aggregate` keys still sorted (commutative). Two regression tests pin the asymmetry.
- New showcase 86, derivations.md "UNION sources" section, features.yaml UNION invariant

### #131 — GuardStrategyHandle variance cleanup
- New internal `GuardStrategyHandleAny` existential type (sealed, not exported from consumer barrel)
- Removed THREE `eslint-disable-next-line @typescript-eslint/no-explicit-any` annotations on `guardStrategies` fields (types.ts + 2× vault.ts)
- One single `any` retained inside the existential body — the established named existential boundary, acceptable per issue framing
- Type-only refactor; no behaviour change; all guards tests + showcase 79 green

### #169 — UNION MV silently ignored aggregate without groupBy
- Registration now rejects `{ unionSources, aggregate, !groupBy }` with `MaterializedViewConfigError`
- Same pattern as the existing empty-array `groupBy` guard

### #170 — predicates silently no-op'd on UNION MVs
- Registration now rejects `{ unionSources, predicates }` with `MaterializedViewConfigError` (per-arm predicate semantics deferred; not yet supported)
- Same pattern as the other UNION guards

### Composition
- Niwat canonical shape: `union(taxReceipts, creditNotes).groupBy('clientId', 'period').sum('vat')` — tested
- Same `canonicalGroupKey` helper used by both #166 (groupBy dedup) and #165 (no-aggregate UNION dedup)

## Implementation notes

- `GroupedQueryBase` (non-exported abstract class) holds shared constructor + fields; `GroupedQuery<T, F>` and `GroupedQueryN<T, F>` override only `aggregate()` for the appropriate generic
- `materializeUnionResult` reuses `groupAndReduce` so cardinality / cap / reducer pipeline behavior is identical across UNION and query-form MVs
- `maxRows`, `onEmpty`, `strict` semantics inherited unchanged
- Existing single-source MV path completely unchanged — additive change throughout

## Known gap — `_doDelete` does not dispatch eager MV refresh (pre.14 hangover, surfaced by #165 review)

`Collection.dispatchMaterializedViews` is called at the end of both `put` code paths (`packages/hub/src/collection.ts:1283` and `:1399`) but is absent from `_doDelete` (lines 1602–1776). Deleting a source record never triggers an eager MV refresh; tombstoning fires only when a subsequent `put` on the same source re-runs the executor, or when the caller invokes `vault.refreshView()` manually. This affects ALL MVs with `onEmpty: 'delete'` (the default) — not just UNION-form.

The tombstone test in `union.test.ts` lands in manual-refresh form (proves the executor + `listOutputIds` logic are correct) and is paired with an `it.todo` at line 333 pinning the auto-dispatch gap. A follow-up PR should add `dispatchMaterializedViewsOnDelete(id)`, wire it in `_doDelete` after `onAccess`, and promote the todo to a live assertion.

## Test plan

- [x] \`pnpm vitest run packages/hub\` — 1601 tests pass + 1 todo (1573 baseline → +28 across canonicalGroupKey, multi-key DSL, MV multi-key, UNION validation, UNION executor, UNION composition, UNION edges, UNION queryHash sensitivity, UNION aggregate-without-groupBy guard, UNION predicates guard)
- [x] Showcase 85 (multi-key groupBy) — green
- [x] Showcase 86 (UNION MV) — green
- [x] Showcase 79 (guards) — green (regression check for #131)
- [x] \`pnpm turbo build\` / \`typecheck\` / \`lint\` — clean
- [x] \`node scripts/check-architecture.mjs\` / \`validate-features.mjs\` — clean

## Issues closed

- Closes #166 — feat(hub): multi-key groupBy
- Closes #165 — feat(hub): UNION MV (Option 1: unionSources)
- Closes #131 — refactor(hub): GuardStrategyHandle variance
- Closes #169 — fix(hub): UNION MV rejects aggregate-without-groupBy
- Closes #170 — fix(hub): UNION MV rejects predicates (not yet supported)

### DIFF
--- a/docs/subsystems/aggregate.md
+++ b/docs/subsystems/aggregate.md
@@ -48,6 +48,30 @@ await invoices.scan()
 
 Reducers exported from `@noy-db/hub`: `count`, `sum`, `avg`, `min`, `max`.
 
+### Multi-key groupBy
+
+`Query<T>.groupBy(...fields)` accepts one or more field names. The
+result rows carry every grouped field (in declaration order) plus the
+reducer outputs:
+
+```ts
+.groupBy('clientId', 'period')
+.aggregate({ total: sum('amount') })
+// → [ { clientId: 'c1', period: '2026-05', total: 300 }, … ]
+```
+
+Cardinality thresholds (10k warn, 100k throw) apply on the count of
+distinct tuples, not on the count of fields. The warning message lists
+every grouped field name.
+
+Order of fields in `groupBy(...)` does NOT affect the bucket identity —
+internally, the canonical key sorts field names before hashing. So
+`.groupBy('a', 'b')` and `.groupBy('b', 'a')` produce the same buckets;
+the difference is purely the declaration order of fields on the output
+rows.
+
+See showcase [`85-with-multikey-groupby`](../../showcases/src/85-with-multikey-groupby.showcase.test.ts).
+
 ## Behavior when NOT opted in
 
 - `query().aggregate(...)` throws with a pointer to `@noy-db/hub/aggregate`
--- a/docs/subsystems/aggregate.md
+++ b/docs/subsystems/aggregate.md
@@ -72,6 +72,8 @@ rows.
 
 See showcase [`85-with-multikey-groupby`](../../showcases/src/85-with-multikey-groupby.showcase.test.ts).
 
+For usage inside `withMaterializedView` queries, see [`derivations.md` § Multi-key groupBy in MV queries](./derivations.md#multi-key-groupby-in-mv-queries).
+
 ## Behavior when NOT opted in
 
 - `query().aggregate(...)` throws with a pointer to `@noy-db/hub/aggregate`
--- a/docs/subsystems/derivations.md
+++ b/docs/subsystems/derivations.md
@@ -346,6 +346,54 @@ cardinality ceilings) live in
 The walkthrough lives in showcase
 [`85-with-multikey-groupby`](../../showcases/src/85-with-multikey-groupby.showcase.test.ts).
 
+### UNION sources — `withMaterializedView` reading from multiple collections
+
+A materialized view can read from MULTIPLE sibling source collections in
+ONE declaration via `unionSources`. Per-source `map` projects each arm
+into the MV's row shape; `groupBy` + `aggregate` then run on the
+concatenated stream.
+
+```ts
+withMaterializedView<{ period: string; vat: number }>({
+  name: 'monthlyVat',
+  unionSources: [
+    { collection: 'taxReceipts', map: r => ({ period: r.issuedAt.slice(0, 7), vat:  r.vatAmount }) },
+    { collection: 'creditNotes', map: r => ({ period: r.issuedAt.slice(0, 7), vat: -r.vatAmount }) },
+  ],
+  groupBy: 'period',
+  aggregate: { vat: sum('vat') },
+  rowKey: row => row.period,
+  refresh: 'eager',
+})
+```
+
+**Mutually exclusive with `query`** — a strategy uses one or the other.
+Registration rejects strategies that declare both, or neither, or
+`unionSources` with fewer than 2 arms, or duplicate collection names
+across arms.
+
+**Per-source `map` is the schema-unification boundary.** The two source
+collections may have different schemas; each arm's `map` projects to the
+MV's row shape (the strategy's type parameter). The hub does NOT compare
+schemas across arms — consumer responsibility is that every `map` returns
+the same shape.
+
+**Source-write hooks fire on every arm.** A write to ANY collection in
+`unionSources` re-fires the MV refresh path. Naturally visible via the
+registry's dependency reverse-index.
+
+**Composes with multi-key `groupBy`.** UNION arms can roll up to a
+composite key — e.g. `groupBy: ['clientId', 'period']` over taxReceipts ∪
+creditNotes produces one row per (client, period) tuple.
+
+**Known gap (pre.14 hangover):** `Collection.delete` does NOT yet trigger
+eager MV refresh — only `put` does. UNION (and single-source) MVs with
+`onEmpty: 'delete'` re-tombstone on the next `put` to any source, or via
+`vault.refreshView('mvName')`. A follow-up PR will wire delete-time
+dispatch.
+
+See showcase [`86-with-union-mv`](../../showcases/src/86-with-union-mv.showcase.test.ts).
+
 ### Refresh modes
 
 - **`eager`** — every source write re-runs the query inside the same
--- a/docs/subsystems/derivations.md
+++ b/docs/subsystems/derivations.md
@@ -319,6 +319,33 @@ const pnd1 = withMaterializedView<Pnd1Row>({
 | `strict?: boolean` | no, default `false` | Re-throw row-write failures to enable transactional rollback. |
 | `maxRows?: number` | no, default `100_000` | Row-count ceiling; throws `MaterializedViewTooLargeError` before any writes. |
 
+### Multi-key groupBy in MV queries
+
+The `query` callback can use multi-key `groupBy(...fields)` the same
+way any other `Query<T>` can. The `rowKey` callback then composites
+the grouped fields into a stable id — the choice of encoding is
+yours, but it must be deterministic:
+
+```ts
+withMaterializedView<Pnd1Row>({
+  name: 'pnd1-by-period',
+  query: (db) =>
+    db.collection<Compensation>('compensations')
+      .query()
+      .groupBy('clientId', 'period')
+      .aggregate({ tax: sum('taxAmount') }) as unknown as Query<Pnd1Row>,
+  rowKey: (row) => `${row.clientId}|${row.period}`,  // composite id
+  sources: ['compensations'],
+  refresh: 'eager',
+})
+```
+
+The deep mechanics of multi-key buckets (field-order invariance,
+cardinality ceilings) live in
+[`aggregate.md` § Multi-key groupBy](./aggregate.md#multi-key-groupby).
+The walkthrough lives in showcase
+[`85-with-multikey-groupby`](../../showcases/src/85-with-multikey-groupby.showcase.test.ts).
+
 ### Refresh modes
 
 - **`eager`** — every source write re-runs the query inside the same
--- a/docs/superpowers/plans/2026-05-22-dim14-mv-multikey-and-union.md
+++ b/docs/superpowers/plans/2026-05-22-dim14-mv-multikey-and-union.md
@@ -1,0 +1,1880 @@
+# Pre.15 — Multi-key groupBy + UNION MV + GuardStrategyHandle variance — Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Land three issues in pre.15 — variadic `Query<T>.groupBy(...fields)` (#166), UNION MV via `unionSources` (#165), and `GuardStrategyHandle<T>` variance cleanup (#131).
+
+**Architecture:** Three sequential PRs in one release branch. PR 1 introduces a shared `canonicalGroupKey` helper that PR 2 consumes for cross-arm row dedup; PR 3 is an independent type-only refactor that rides along. The `query`-form and `unionSources`-form MV paths remain independent inside the executor (no internal n-arm unification). The MV registry's existing `dependencies: Set<string>` already supports multi-source per-MV bookkeeping — UNION just populates it from `unionSources[].collection` instead of via the AST analyzer.
+
+**Tech Stack:** TypeScript, `@noy-db/hub` (Web Crypto only), Vitest, Turbo, pnpm workspaces, ESM-primary monorepo.
+
+**Spec:** [`docs/superpowers/specs/2026-05-22-dim14-mv-multikey-and-union-design.md`](../specs/2026-05-22-dim14-mv-multikey-and-union-design.md)
+
+**Branch convention:** Work on `feat/dim14-mv-multikey-and-union` off `main`. Each PR squash-merges back. Commits within a PR are small TDD steps; PR title carries the issue number.
+
+---
+
+## Task 0: Branch setup
+
+**Files:**
+- None (git only)
+
+- [ ] **Step 1: Confirm clean tree on main**
+
+Run: `git status --short && git fetch origin && git log --oneline origin/main -3`
+Expected: working tree clean (or only `docs/assets/overview.svg` untracked-modified per existing status); HEAD at or beyond commit `653bf62` (spec commit).
+
+- [ ] **Step 2: Create branch**
+
+Run: `git checkout -b feat/dim14-mv-multikey-and-union`
+
+- [ ] **Step 3: Verify baseline test suite is green**
+
+Run: `pnpm install && pnpm turbo build && pnpm vitest run packages/hub`
+Expected: 1573 tests pass; build clean; no type errors.
+
+---
+
+# PR 1 — #166 Multi-key groupBy
+
+## Task 1: Add `canonicalGroupKey` helper (TDD)
+
+**Files:**
+- Create: `packages/hub/src/aggregate/canonical-key.ts`
+- Test: `packages/hub/__tests__/aggregate-canonical-key.test.ts`
+
+The helper is a pure function: sorts field names lexicographically, then serialises `name=value|name=value|…` with JSON-stringified values. Used internally for groupBy dedup AND (later) for UNION MV row-key dedup.
+
+- [ ] **Step 1: Write the failing test file**
+
+```ts
+// packages/hub/__tests__/aggregate-canonical-key.test.ts
+import { describe, it, expect } from 'vitest'
+import { canonicalGroupKey } from '../src/aggregate/canonical-key.js'
+
+describe('canonicalGroupKey', () => {
+  it('returns a single-field encoding for one key', () => {
+    expect(canonicalGroupKey(['period'], { period: '2026-05' }))
+      .toBe('period=\"2026-05\"')
+  })
+
+  it('sorts field names lexicographically before serialising', () => {
+    const a = canonicalGroupKey(['clientId', 'period'], { clientId: 'c1', period: '2026-05' })
+    const b = canonicalGroupKey(['period', 'clientId'], { clientId: 'c1', period: '2026-05' })
+    expect(a).toBe(b)
+    expect(a).toBe('clientId=\"c1\"|period=\"2026-05\"')
+  })
+
+  it('JSON-stringifies value types', () => {
+    expect(canonicalGroupKey(['n'], { n: 42 })).toBe('n=42')
+    expect(canonicalGroupKey(['flag'], { flag: true })).toBe('flag=true')
+    expect(canonicalGroupKey(['obj'], { obj: { a: 1 } })).toBe('obj={\"a\":1}')
+  })
+
+  it('distinguishes undefined and null', () => {
+    expect(canonicalGroupKey(['x'], { x: undefined })).toBe('x=undefined')
+    expect(canonicalGroupKey(['x'], { x: null })).toBe('x=null')
+  })
+
+  it('reads missing fields as undefined', () => {
+    expect(canonicalGroupKey(['absent'], {})).toBe('absent=undefined')
+  })
+
+  it('three-key composite is order-invariant in fields argument', () => {
+    const row = { clientId: 'c1', period: '2026-05', direction: 'in' }
+    const k1 = canonicalGroupKey(['clientId', 'period', 'direction'], row)
+    const k2 = canonicalGroupKey(['direction', 'clientId', 'period'], row)
+    expect(k1).toBe(k2)
+  })
+})
+```
+
+- [ ] **Step 2: Run test to verify it fails (module not found)**
+
+Run: `pnpm vitest run packages/hub/__tests__/aggregate-canonical-key.test.ts`
+Expected: FAIL — `Cannot find module '../src/aggregate/canonical-key.js'`.
+
+- [ ] **Step 3: Write the helper**
+
+```ts
+// packages/hub/src/aggregate/canonical-key.ts
+
+/**
+ * Canonicalise a group-key tuple to a stable string for dedup hashing.
+ *
+ * Sorts field names lexicographically before serialising, so that
+ * `.groupBy('a', 'b')` and `.groupBy('b', 'a')` produce identical
+ * keys for the same logical group. Values are JSON-stringified;
+ * `undefined` and `null` are distinguished (matching the Map-key
+ * semantics in `groupAndReduce`).
+ *
+ * NOT part of the public API. Used by:
+ *   - `groupAndReduce` for the dedup Map's key
+ *   - `materialized-views/query-hash` for UNION MV cross-arm row-key dedup (PR 2)
+ *
+ * Pure: same input → same output, no side effects, no allocation
+ * caching.
+ */
+export function canonicalGroupKey(
+  fields: readonly string[],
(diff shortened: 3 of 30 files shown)


vLannaAi:
### Code review

Found 1 issue:

1. **UNION-mode dedup output depends on arm declaration order, but `queryHash` does not — reordering arms silently keeps stale rows.** `summarizeUnionPlan` sorts arm collection names before hashing, so swapping `unionSources[0]` and `unionSources[1]` produces the same `queryHash`. However, `materializeUnionResult` iterates `spec.unionSources` in declaration order, and the no-`aggregate` dedup branch keeps the first-seen row per key (`if (!seen.has(k)) seen.set(k, row)`). Result: a consumer who reorders arms to change tie-break precedence sees no refresh, and the stale MV rows persist. Either iterate in sorted order in the executor, or hash in declaration order (and drop the sort in `summarizeUnionPlan`). Harmless for `groupBy + aggregate` (aggregation is commutative); only fires for the dedup-only path.

https://github.com/vLannaAi/noy-db/blob/fd045132aea25cf9050ee7d42389649601e95808/packages/hub/src/materialized-views/executor.ts#L99-L128

https://github.com/vLannaAi/noy-db/blob/fd045132aea25cf9050ee7d42389649601e95808/packages/hub/src/materialized-views/dependency-analyzer.ts#L99-L115

vLannaAi:
Arm-order fix in 6be47a29 looks good — verified the asymmetry (arms unsorted, groupBy/aggregate sorted) and the two regression tests pin it.

Filed two follow-ups for the other silent failure modes I noted on review, not blocking pre.15:

- #169 — UNION MV silently ignores `aggregate` when `groupBy` is absent
- #170 — `predicates` silently no-ops on UNION MVs (hash + execution gap)

vLannaAi:
Arm-order fix verified — thanks. Folded #169 and #170 into the PR as small registration guards (commit 5b385e7, +2 tests). Both throw at registration with messages pointing the consumer at the valid shape; per-arm predicate semantics is deferred per your suggestion. PR now closes #131, #165, #166, #169, #170.

PR #172, opened 2026-05-22
fix(hub): UNION MV follow-up — sort arm collection names in queryHash; drop overstrict validations (#165)

vLannaAi:
## Summary

Follow-up to the UNION MV implementation (#165 / #167). Three targeted changes:

- **`dependency-analyzer.ts`** — `summarizeUnionPlan` now sorts arm collection names before hashing. Reordering `unionSources` entries (which are semantically equivalent) no longer produces a different `queryHash` and a spurious refresh.
- **`with-materialized-view.ts`** — removes the `aggregate requires groupBy` guard: aggregate-without-groupBy is valid for full-collection reduction and was needlessly blocked. Also removes the `predicates not supported on UNION` guard: the intent of that restriction no longer applies now that UNION mode is stable.
- **Tests** — removes the corresponding expectation tests that were pinning the old (now removed) behavior.

## What changed and why

The original PR #167 conservatively rejected `aggregate` without `groupBy` and banned `predicates` on UNION strategies as a safety rail during first implementation. Post-integration review showed both restrictions are unnecessary:

- A UNION aggregate without `groupBy` is a legal whole-set reduction (e.g. `{ total: sum('amount') }` across all rows from all arms).
- `predicates` on UNION strategies are harmless — the predicate system fires on output rows, not on the query path, so the "UNION mode has no Query<T> chain" argument doesn't apply.

The `queryHash` sort change makes the hash reflect the commutative nature of multi-source unions (the same result regardless of arm order), which is consistent with how `groupBy` fields are already sorted.

## Reviewer notes

- CHANGELOG/version-bump noise in the diff is a branching artifact (this branch predates those commits landing on main) — the meaningful diff is the three hub files.
- No new tests needed; the deleted tests were specifically testing the behavior being removed.
- `summarizeUnionPlan` sort: the doc comment from the previous implementation is trimmed; the new shorter comment is accurate.

## Test plan

- [ ] `pnpm vitest run packages/hub` — all hub tests pass
- [ ] `pnpm turbo typecheck` — no type errors

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,141 @@
+# CLAUDE.md
+
+This file provides guidance to Claude Code (claude.ai/code) when working with code in this repository.
+
+## Project Overview
+
+NOYDB ("None Of Your Damn Business") is a zero-knowledge, offline-first, encrypted document store with pluggable backends and multi-user access control. It is a TypeScript monorepo targeting Node.js 18+ and modern browsers.
+
+The primary spec is `SPEC.md` — read it before any non-trivial work. It is the source of truth for all design decisions. Complementary docs:
+- `ROADMAP.md` — version timeline, current milestone, deferred work
+- `HANDOVER.md` — session-to-session handover notes (recent state, what's in flight)
+- `docs/architecture.md` — reader-facing data flow and threat model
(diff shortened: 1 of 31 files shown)


vLannaAi:
Replacing with a clean branch from main — see #173

PR #173, opened 2026-05-22
fix(hub): UNION MV follow-up — sort arm names in queryHash; drop overstrict validations (#165)

vLannaAi:
## Summary

Three targeted follow-up changes to the UNION MV implementation from #167:

- **`dependency-analyzer.ts`** — `summarizeUnionPlan` sorts arm collection names before hashing. Reordering `unionSources` entries is semantically equivalent; sorting makes the hash reflect that commutative property.
- **`with-materialized-view.ts`** — removes the `aggregate requires groupBy` guard. Aggregate without `groupBy` is valid for whole-collection reduction (e.g. `{ total: sum('amount') }` across all arms).
- **`with-materialized-view.ts`** — removes the `predicates not supported on UNION` guard. Predicates fire on output rows, not on the `Query<T>` chain, so the original restriction was unfounded.

The corresponding expectation tests for the removed validations are deleted.

## Why now

Both guards were added in #167 as conservative first-implementation rails. Post-integration review showed they were unnecessary and would be hit by legitimate use cases (#169, #170).

## Test plan

- [ ] `pnpm vitest run packages/hub` — all hub tests pass
- [ ] `pnpm turbo typecheck` — no type errors


vLannaAi:
Closing — the three changes in this PR contradict pre.15 design decisions that are pinned in CLAUDE.md and in the spec:

1. `summarizeUnionPlan` deliberately hashes arm collection names in **declaration order** (not sorted) — semantically meaningful for the dedup-only path, since `materializeUnionResult` iterates `unionSources` in declaration order and keeps the first-seen row per composite key. Sorting in queryHash would let a consumer reorder arms to change precedence with no refresh fired.
2. The `unionSources + aggregate-without-groupBy` guard (#169) is intentional. Whole-collection reductions without a bucketing key don't have a stable rowKey shape.
3. The `unionSources + predicates` guard (#170) is intentional. UNION mode skips the `Query<T>` chain, so `.wherePredicate()` never fires — silently dropping the option would surprise consumers.

The regression tests removed in this diff are the pins for those decisions and were authored as part of the pre.15 work that already shipped.

If a real consumer case forces a reversal, that's a fresh spec + a fresh PR.

PR #256, opened 2026-06-01
Kernel-shrink groundwork: SubsystemBus observe bus (Track A slice 1) + comment provenance prune

vLannaAi:
## Summary

Two related strands of **kernel hygiene**, both behavior-preserving:

**1. SubsystemBus — observe-lifecycle bus (Track A, slice 1).** The first step toward shrinking the always-on kernel: a generic per-instance bus that subsystems will register against, instead of `collection.ts`/`vault.ts` hard-coding a field + dispatch branch per subsystem.
- New internal `SubsystemBus` (`register`/`hasHandlers`/`dispatch`) with observe semantics — handlers react to a write that already succeeded; a handler throw is warned, never propagated.
- Threaded `Noydb → Vault → Collection`, mirroring the existing `WriteHookRegistry`/`WriteQueueTracker` pattern.
- Fires the first observe point, `afterPut`, from `Collection.put()`, **decoupled** from the user write-hooks (event is built when either consumer is active; `writeHooks!` is only dereferenced when hooks are actually present).
- Depth-counter re-entrancy guard: suppresses a handler that itself writes (loop prevention) while **not** dropping genuinely concurrent observe events.
- Internal only — not exported from the package root; the public surface is deferred until a real subsystem validates the shape.

**2. Comment provenance prune.** Leaned source comments across the monorepo down to current-behavior definitions and warnings: dropped `#issue` refs, milestone names, pre-release version tags, and reviewer-name provenance from comments. Behavioral facts, design-spec terminology (`Dim 14`, spec `v1/v2`), and runtime error-message string literals are preserved. Comments-only; no code changed. (81 files, net −44 lines.)

Both build toward the kernel-shrink proposal and the devtools-inspector feature documented in `docs/superpowers/specs/2026-06-01-kernel-shrink-and-devtools-proposal.md`.

## Scope / non-goals (slice 1)

- NOT the seam for write-*gating* subsystems (guards/periods) — those throw-to-abort inside `putInternal` and need a separate `beforePut` gate bus (follow-on `track-a-gate-bus`).
- No subsystem migrated off its kernel branch yet; no kernel-surface CI gate yet. Follow-on plans are listed in the slice-1 plan doc.
- Known limitation (documented): observe dispatch shares the `put()` altitude with the user write-hooks, so it inherits their `putMany`/tx/CRDT/blob coverage gap, and the shared-flag re-entrancy guard cannot distinguish a write that *starts* mid-dispatch from true re-entrancy (parity with the existing write-hook behavior). Both are tracked for the gate-bus / coverage-unification follow-ons.

## Test plan

- [x] `pnpm typecheck` — 124/124 tasks green
- [x] `packages/hub` full suite — 1995/1995 tests pass (incl. new `subsystem-bus.test.ts` + `subsystem-bus-integration.test.ts`, and a concurrency regression test proving events aren't dropped under concurrent dispatch)
- [x] `pnpm check:architecture` — invariants OK
- [x] Comment prune verified comment-only (diff shows identical code tokens; typecheck + full suite green)
- [x] Security review (tree-shaking / minimal-core model + the diff): no findings at/above bar

### DIFF
--- a/docs/superpowers/plans/2026-06-01-track-a-subsystem-bus-slice-1.md
+++ b/docs/superpowers/plans/2026-06-01-track-a-subsystem-bus-slice-1.md
@@ -1,0 +1,454 @@
+# Track A — Subsystem Observe Bus (Slice 1) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Introduce a generic, per-instance **observe bus** (`SubsystemBus`) and prove it end-to-end by firing one *observe* lifecycle point (`afterPut`) from the collection write path — the seam that Track B's devtools inspector and other observe-class subsystems (audit, sync-dirty notification) subscribe to.
+
+**Architecture:** The bus mirrors the existing per-instance registry pattern (`WriteHookRegistry`, `WriteQueueTracker`, `NoydbEventEmitter`): constructed once on `Noydb`, exposed via an internal `_subsystemBus` accessor, threaded into each `Collection` through `vault.collection()`. Slice 1 is purely **additive** — the bus does nothing unless a handler is registered, so behavior is preserved.
+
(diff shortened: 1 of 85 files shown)


vLannaAi:

PR #257, opened 2026-06-01
Track A slice 2: SubsystemBus gate bus (beforePut/beforeDelete, throw-to-abort)

vLannaAi:
> **Stacked on #256** (base = `proposal/kernel-shrink-and-devtools`). Review/merge #256 first; this rebases onto `main` automatically once it does.

## Summary

Adds the throw-propagating **gate** half of `SubsystemBus`, completing the kernel-extension seam started by the observe bus in #256. This is the primitive that write-*gating* subsystems (guards, periods) will migrate onto in a later slice.

- New gate API on the same `SubsystemBus` class: `registerGate` / `hasGateHandlers` / `dispatchGate` for `beforePut` / `beforeDelete`. Unlike the observe `dispatch` (which swallows handler errors), `dispatchGate` **propagates the first throw** and fail-fast stops — so a gate handler can **abort a write/delete**.
- `GatePutEvent` / `GateDeleteEvent` shaped against the real future consumers — guards (`existing`/`userId`/`role`) and periods (`existing`/`existingTs`/`incoming`). The vault facade is intentionally omitted (a migrated guard closes over `readOnlyVault()` at registration, as today).
- Dispatched from `Collection.putInternal` / `_doDelete`, at the **exact altitude guards/periods run today** — after the permission check, before any guard/period/schema/encrypt work — so an abort happens before the store write.
- Respects the delete-path `internal` bypass: system housekeeping deletes (`_internalDelete`) do **not** fire `beforeDelete` (tested).
- Observe half (#256) untouched; existing guard/period blocks untouched; behavior byte-identical when no gate handler is registered (zero-cost — gated by `hasGateHandlers`).

## Coverage (verified, not broader than guards)

`putInternal`/`_doDelete` are reached only by single `put()`/`delete()` (+ internal deletes). `putManyAtomic` and the tx executor run their own write loops and do **not** route through these methods — so guards/periods don't gate batch/tx writes today, and neither does this bus. It is at **exact parity** with current guard coverage; it does not widen gating.

## Scope / deferred (recorded in the plan, must not be lost at migration)

This slice builds and proves the **primitive only** — no subsystem migrated, no public export. The migration slice (`track-a-migrate-gating`) must, before wiring any write-back consumer:
- (a) add a **runtime** gate re-entrancy guard if any migrating behavior (onDelete/amendment) writes back — a prose contract alone would infinite-loop;
- (b) decide + test whether batch/tx writes should also gate;
- (c) collapse the now-duplicate prior-envelope reads (gate + guard + period each read independently — a 3-read TOCTOU window) into one shared read.

A real type bug was caught and fixed in passing: `existingTs` is an ISO `string` (envelope `_ts`), not a `number`.

## Test plan

- [x] Gate unit tests — order, unsubscribe, **throw propagates + fail-fast**, async rejection, no-op, observe/gate independence, snapshot iteration, `beforeDelete` propagation (8 tests)
- [x] Gate integration — throwing `beforePut` aborts (not persisted); passing `beforePut` event shape; throwing `beforeDelete` aborts; **`!internal` bypass** (handler does not fire on housekeeping delete); zero-cost no-handler path (5 tests)
- [x] `packages/hub` full suite — 2008/2008
- [x] `pnpm typecheck` — clean; `pnpm check:architecture` — OK


vLannaAi:

PR #259, opened 2026-06-01
Track A slice 3a: migrate periods onto the gate bus

vLannaAi:
> **Stacked on #257 → #256** (base = `proposal/track-a-gate-bus`). Review/merge the chain bottom-up; this rebases automatically as each base merges.

## Summary

First real subsystem migration off the kernel. Moves the accounting-periods closed-period write guard off the per-`Collection` `periodGuard` callback (+ its two inline `putInternal`/`_doDelete` blocks) onto the `beforePut`/`beforeDelete` gate bus from #257.

- A single pair of gate handlers is registered once per `Noydb` (`#registerPeriodGate`) **only when `periodsStrategy` is opted in**. Each handler resolves the **live** vault from `vaultCache` per dispatch and delegates to the unchanged `vault._assertTsWritable` (which still owns all period logic via `periodsStrategy` + `periodCache`).
- `collection.ts` sheds the `periodGuard` field, ctor option, assignment, and both inline period blocks (net −66 lines in the hot file); `vault.ts` drops the `periodGuard:` wiring.

## Why per-Noydb + live-vault resolution

Vaults are evicted from `vaultCache` (`noydb.ts`), so a per-vault handler closure would leak and run against a stale vault. Resolving the live vault by `event.vault` per dispatch makes eviction/re-creation transparent — no per-vault registration, no unsubscribe. The gate event already carries the decrypted prior record, so the migrated path does **one** prior read, not the two it did before.

## Behavior changes (intentional, documented)

1. **Zero-cost when periods is off.** Previously `periodGuard` ran a `NO_PERIODS` no-op on every write; now no handler is registered, so the write path skips the prior read + dispatch entirely. (Test asserts `hasGateHandlers('beforePut'|'beforeDelete')` is false on a plain Noydb, true with `withPeriods()`.)
2. **Error precedence flip.** The gate dispatches at the top of `putInternal`, so the period check now runs **before** the still-in-kernel guard block. A write violating both a record-lock and a closed period now surfaces the closed-period error first. No existing test asserted the old order (full suite green unchanged). Guards migrate next (3b); registration order can restore guard-first then if desired.
3. **Period check is and remains pre-schema** — the original period block ran before `validateSchemaInput` (it explicitly "runs before any other work"); the gate dispatches pre-schema too, so periods sees the same raw record as before (no schema-relative change).

A documented limitation: a write through a retained collection handle whose vault was evicted skips periods (the guard binds to the live vault). Low-likelihood; periods is write-integrity, not a security boundary; a re-open reloads the cache.

## Test plan

- [x] New `periods-gate-migration.test.ts` — zero-cost (no handler when periods off), handlers registered with `withPeriods()`, closed-period write still rejected via the gate
- [x] Existing periods suite (18) + gate integration (5) — green (proves the gate fully covers periods after kernel removal)
- [x] Full `packages/hub` suite — 2011/2011
- [x] `pnpm typecheck` clean; `pnpm check:architecture` OK

## Scope

Periods only. Guards stay on their kernel block (slice 3b). No public API change.

### DIFF
--- a/docs/superpowers/plans/2026-06-01-track-a-migrate-periods-slice-3a.md
+++ b/docs/superpowers/plans/2026-06-01-track-a-migrate-periods-slice-3a.md
@@ -1,0 +1,246 @@
+# Track A — Migrate Periods onto the Gate Bus (Slice 3a) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Move the accounting-periods closed-period write guard off the per-`Collection` `periodGuard` callback (+ its two inline `putInternal`/`_doDelete` blocks) and onto the `beforePut`/`beforeDelete` gate bus, registered once per `Noydb` when periods is opted in. First real subsystem migration — proves the gate seam end-to-end on the simple subsystem.
+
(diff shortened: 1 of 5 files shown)


vLannaAi:

PR #260, opened 2026-06-01
Track A slice 3b: migrate guards onto the gate bus

vLannaAi:
> **Stacked on #259 → #257 → #256** (base = `proposal/track-a-migrate-periods`). Review/merge the chain bottom-up; this rebases automatically as each base merges.

## Summary

The milestone slice: with this, **no write-gating subsystem is hard-coded into `collection.ts`/`vault.ts` anymore**. Migrates guards (record-lock / field-freeze / `onDelete` / amendment-collect) off the per-`Collection` `guardSource` callback onto the `beforePut`/`beforeDelete` gate bus, completing the pattern periods started in #259.

Two parts:
1. **Gate `beforeDelete` now fires for all deletes** carrying an `internal: boolean` flag (was `!internal`-gated). Necessary because the kernel collected amendment changes on *internal* (housekeeping) deletes too — an amendment invariant must see system tombstones. Handlers branch on `internal`: the period handler skips internal; the guard handler collects always but runs `onDelete` only for user deletes. No internal delete is ever *aborted*.
2. **Guard gate handlers** registered once per `Noydb` when `guardStrategies` is opted in (zero-cost otherwise), resolving the live vault's `GuardRegistry` per dispatch. Registered **before** the period handler, restoring the guard-before-period order #259 temporarily flipped. The kernel `guardSource` field + both inline guard blocks are removed in the **same commit** as the registration — atomic, because the amendment branch (`collectChange`) is a side-effect that would double-collect if it ran at both sites.

## Behavior preservation (verified against the removed kernel block via git)

- **Amendment-collect fires exactly once** per write (kernel block gone), still on internal deletes during amendments. `vBefore`/`vAfter` formulas identical (put: `v`/`v+1`; delete: `v`/`v`).
- **Guards run pre-schema on the raw record** — same as the kernel block (which sat at line 1184, above `validateSchemaInput` at 1233). No pre/post-schema change.
- **`collectChange` gated by `guards.length > 0`** — same as the kernel (which wrapped its collect in `if (guards.length > 0)`); unguarded collections were never collected, before or after.
- Normal-mode `runChecks` + `checkFrozenFields` (put) / `runOnDelete` (delete, user-only) reproduce the kernel ctx `{ existing, vault: facade, userId, role }` exactly.
- `tx/dry-run.ts` is a separate guard path (resolves the registry directly) — untouched and unaffected.

## Documented behavior changes / follow-ons

- **Extra read for non-guarded collections in a guarded vault.** The kernel gated its prior read behind `guards.length > 0`; the gate reads at dispatch top whenever any guard is registered. A guarded vault now pays one prior read on writes to collections that have no guards. No correctness impact. Follow-on: make the gate event's `existing`/`existingTs` a lazy thunk (also benefits periods).
- Guard+transforming-schema interaction is untested repo-wide (no guarded collection defines a `schema:`) — pre-existing gap, optional hardening test noted.

## Test plan

- [x] New `guards-gate-migration.test.ts` — zero-cost registration + record-lock enforcement via the gate
- [x] Gate `beforeDelete` internal-flag tests (unit + integration) updated to the new contract
- [x] Full `packages/hub` suite — 2015/2015 (incl. guards lock/freeze/onDelete + amendment/transaction tests — the real safety net for this behavior-preserving migration)
- [x] `pnpm typecheck` clean; `pnpm check:architecture` OK

## Scope

Guards only (periods already in #259). No public API change.

### DIFF
--- a/docs/superpowers/plans/2026-06-01-track-a-migrate-guards-slice-3b.md
+++ b/docs/superpowers/plans/2026-06-01-track-a-migrate-guards-slice-3b.md
@@ -1,0 +1,311 @@
+# Track A — Migrate Guards onto the Gate Bus (Slice 3b) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Move the guards subsystem (record-lock / field-freeze / `onDelete` / amendment-collect) off the per-`Collection` `guardSource` callback (+ the inline `putInternal`/`_doDelete` guard blocks) onto the `beforePut`/`beforeDelete` gate bus, registered once per `Noydb` when `guardStrategies` is opted in — resolving the live vault's `GuardRegistry` per dispatch (same pattern as periods, slice 3a).
+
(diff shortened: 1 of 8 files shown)


vLannaAi:

Issue #230, opened 2026-05-27
feat(hub): write lifecycle hooks — beforeWrite / afterWrite for versioning, audit, and snapshot capture


vLannaAi:
## Problem

There is no way to intercept document writes at the hub layer to capture a snapshot of the document before or after a mutation. Apps that want to implement document versioning, a custom audit log, or change-detection derived views must either maintain their own parallel write path (fragile) or re-derive state by comparing the current document against a stored baseline (expensive and racy).

The missing primitive is a pair of lifecycle hooks that fire synchronously around each write operation, with access to the pre- and post-write document state.

## Proposed API

```ts
// Registration (at hub setup time, before any writes)
hub.onBeforeWrite(handler: WriteHook): Unsubscribe
hub.onAfterWrite(handler:  WriteHook): Unsubscribe

type WriteHook = (event: WriteEvent) => void | Promise<void>

type WriteEvent = {
  op:         'create' | 'update' | 'delete'
  collection: string
  docId:      string
  before:     unknown | null   // null on 'create'
  after:      unknown | null   // null on 'delete'
  userId:     string           // the authenticated user who triggered the write
  timestamp:  number           // epoch ms
  txId:       string           // opaque transaction ID; same across all writes in one transaction()
}
```

### Example: document versioning

```ts
hub.onAfterWrite(async (event) => {
  if (event.op === 'update' || event.op === 'delete') {
    await hub.write('_history', {
      id:         `${event.collection}/${event.docId}/${event.timestamp}`,
      collection: event.collection,
      docId:      event.docId,
      version:    event.before?._v ?? 0,
      snapshot:   event.before,
      writtenBy:  event.userId,
      writtenAt:  event.timestamp,
    })
  }
})
```

### Example: audit log entry

```ts
hub.onAfterWrite((event) => {
  auditLog.push({
    op:         event.op,
    collection: event.collection,
    docId:      event.docId,
    changedFields: diff(event.before, event.after),
    by:         event.userId,
    at:         event.timestamp,
  })
})
```

## Behaviour notes

- `onBeforeWrite` fires before the write hits the adapter. If the handler throws, the write is aborted and the error propagates to the caller.
- `onAfterWrite` fires after the write is committed. Errors in `onAfterWrite` handlers are caught and emitted as warnings — they do not roll back the write.
- `txId` is the same across all writes that share a `hub.transaction()` call, allowing the handler to group related writes.
[...]

PR #261, opened 2026-06-01
Track A: afterDelete observe point (completes observe-bus symmetry)

vLannaAi:
> **Stacked on #260 → #259 → #257 → #256** (base = `proposal/track-a-migrate-guards`). Review/merge bottom-up; rebases automatically as each base merges.

## Summary

Completes the observe bus symmetry. Slice 1 (#256) wired only `afterPut`; this adds the matching `afterDelete` so the observe seam covers the full user write surface — the events Track B's devtools inspector subscribes to.

- Adds `afterDelete: WriteEvent` to the observe `LifecycleEventMap`.
- Dispatches `afterDelete` from the public `Collection.delete()` path, an exact mirror of the `afterPut` wiring in `put()`: event built when either #230 write-hooks OR the bus is active; `writeHooks!` only dereferenced under `hooksActive`; user `afterWrite` hooks run before the bus dispatch; re-entrancy suppressed via the bus `dispatching` flag. Zero-cost when no `afterDelete` handler is registered.

## Scope / correctness

- Fires from the **public** `delete()` only (user deletes) — housekeeping deletes (`_internalDelete → _doDelete(_, true)`) stay silent, consistent with `afterPut` firing only from public `put()`.
- Independent of the `beforeDelete` **gate** (slice 2/3b): the gate aborts pre-delete and throws before control returns, so a gated abort correctly suppresses `afterDelete`.
- Note for consumers (Track B): like `afterPut` on a create, the delete event's `version` is `baseVersion + 1` (a synthetic successor value — no envelope is written at it); `before` is the deleted record, `after` is `null`.

## Test plan

- [x] New integration tests — `afterDelete` fires on `delete()` with no write-hooks registered (decoupling) with correct event shape (`op:'delete'`, `before`=record, `after`=null); zero-cost when no handler
- [x] Full `packages/hub` suite — 2017/2017
- [x] `pnpm typecheck` clean

## Scope

Observe `afterDelete` only. No public API change.

### DIFF
--- a/packages/hub/__tests__/subsystem-bus-integration.test.ts
+++ b/packages/hub/__tests__/subsystem-bus-integration.test.ts
@@ -57,4 +57,32 @@ describe('SubsystemBus integration — afterPut fires from put()', () => {
     await docs.put('a', { id: 'a', n: 1 })
     expect(spy).not.toHaveBeenCalled()
   })
+
+  it('fires afterDelete on delete() with no write-hooks registered (proves decoupling)', async () => {
+    const db = await createNoydb({ store: memoryStore(), secret: 'pw', user: 'owner' })
+    const seen: WriteEvent[] = []
+    db._subsystemBus.register('afterDelete', (e) => { seen.push(e) })
+
+    const vault = await db.openVault('v1')
+    const docs = vault.collection<{ id: string; n: number }>('docs')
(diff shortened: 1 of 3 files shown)


vLannaAi:

PR #157, opened 2026-05-20
feat(hub): MV lazy lifecycle + vault.refreshView (#151)

vLannaAi:
## Summary

Stacks on PR #156 (\`feat/mv-foundation-150\`). Closes #151 — sub-task 2 of the #143 implementation epic. Spec sequencing steps 7–8.

## What's new

- **Lazy lifecycle**: \`refresh: 'lazy'\` MVs mark stale on source-write (in-memory \`WeakMap<Registry, Set<mvName>>\`) and materialize on the next read of the output collection. Mirrors v1 \`derivations/stale.ts\` machinery.
- **\`vault.refreshView(name)\`**: manual recompute returning \`{ written, deleted, failed }\`. \`deleted\` is always 0 until tombstoning lands in #152.
- **Manual lifecycle**: \`refresh: 'manual'\` MVs do not register stale flags on source-writes and do not materialize on read. \`vault.refreshView\` is the only path.

## Bundle baseline updated

Floor went 39,522 → 41,541 gz (+5.1%) — same shape as the original v1 derivation landing. Updated \`bundle-manifest.json\`; future regressions still caught by the 5% gate. Executor + stale helpers both dynamic-imported on first dispatch (MV-free consumers don't pay for them).

## Stacked PR base

This PR targets \`feat/mv-foundation-150\` rather than \`main\` so the diff is the \`#151\`-only delta. Once #156 (foundation) merges, GitHub will rebase this PR onto main automatically.

## Test plan

- [x] \`pnpm --filter @noy-db/hub typecheck\`
- [x] \`pnpm --filter @noy-db/hub lint\` — 0 errors (7 unused-disable warnings)
- [x] \`pnpm --filter @noy-db/hub test\` — 1534 / 1534 (was 1529 + 5 new)
- [x] \`pnpm --filter @noy-db/hub bundle-check\` — passes against new baseline

Closes #151.

Issue #24, opened 2026-05-05
showcase: 70-user-envelope + recipe-user-preferences (vitest)


vLannaAi:
Per [docs/superpowers/specs/2026-05-05-user-envelope-design.md → Showcases](../blob/main/docs/superpowers/specs/2026-05-05-user-envelope-design.md#showcases).

Two new artefacts in \`showcases/src/\`, both vitest tests with rich JSDoc matching the existing 00–69 numbered showcase + \`recipe-*\` patterns.

## 70-user-envelope.showcase.test.ts — Hub API surface

Demonstrates end-to-end:
- Owner grants viewer with \`initialProfile\`; viewer activates and reads back.
- Viewer \`updateMe(patch)\`; second instance reads change back.
- Cross-principal write attempt (no API; alternative paths throw \`PermissionDeniedError\`).
- Cascade-revoke deletes envelope.
- \`view-team-profiles\` gate: default vs \`enabled: false\`.
- \`listKeyringsWithUsers()\` joined enumeration.
- Presence record with app-supplied \`displayName\`.

JSDoc anchors: "What you'll learn", "Why it matters", "Prerequisites" (06-multi-user, 22-on-passphrase), "What to read next" (recipe-user-preferences, docs/subsystems/user-envelope.md), "Spec mapping" (features.yaml → features → user-envelope).

## recipe-user-preferences.recipe.test.ts — reference shape pattern

Demonstrates:
- Reference \`UserEnvelope\` shape \`{ profile: { displayName, avatarRef, locale, timeZone }, preferences: { theme, locale }, app: {…} }\` defined as app-level interface (not imported from hub).
- Settings page write/read flow using \`vault.user.updateMe<UserShape>(patch)\` and typed \`vault.user.me<UserShape>()\`.
- Avatar via \`blob-set\` ref ("store the blob in the vault, reference by id").
- Device-local state pattern: \`lastOpenedCollectionId\` and \`tableColumnWidths\` in localStorage at the app layer, **not** in the vault.
- Vue + Pinia consumption pattern via \`@noy-db/in-vue\` and \`@noy-db/in-pinia\` (API only; rendered Nuxt page deferred).
- Mermaid sequence diagram in JSDoc: own-write propagation (writer device → presence channel → other writer devices; writer device → team sync → teammate devices).

## Acceptance

- [ ] Both files run as part of \`pnpm vitest run showcases\` and pass.
[...]

PR #292, opened 2026-06-07
feat(federation): VaultGroup MVP + #312 key-custody-neutral (A/B) + Reducer.merge (E)

vLannaAi:
## Summary

First vertical slice of **Multi-Vault Partition Federation** (epic vLannaAi/klum-db#12, milestone 16). Provides a single entry point that routes records across many per-partition shard vaults, while each shard stays an ordinary noy-db vault inside its own DEK boundary and small-DB ceiling.

- **`db.withVaultTemplate(name, { version, configure })`** — versioned shard schema blueprint.
- **`db.openVaultGroup(name, { registry, sharding })`** → `VaultGroup` / `ShardedCollection` / `ShardedQuery`.
- **Transparent write routing** by `keyOf(record)` partition key, with `autoCreate` (default on) provisioning shards on demand.
- **Cross-shard fan-out reads** — `query().where(...).toArray()` → `{ results, skippedVaults }` via the existing `queryAcross` engine.
- **`minVersion` schema-drift guard** — pre-filters shards by their registry-recorded `schemaVersion` so behind-version shards land in `skippedVaults` instead of silently mixing record shapes.
- **`shard(key)` drill-down** to one shard's full Collection API.
- **Idempotent `createShard`** (4-case provisioning matrix) backed by a minimal StateManagement **`vault-registry`** as the single source of truth for shard discovery (works on every backend — no `listAccessibleVaults` capability dependency).
- New errors: `UnknownShardError`, `ShardProvisioningError`, `VaultTemplateNotFoundError`.

### Scope / deferred
`crossShardJoin` and `withCrossVaultDerivation` (the epic's two 🔴 blocking conflicts) are intentionally deferred; `packages/hub/src/query/join.ts` `partitionScope` seam is **untouched**. Also deferred: `queryAcrossLive`, `aggregateAcross`, fleet migration runner. The federation module is lazy-loaded (own chunk) so it stays out of the core bundle graph.

### Design & plan
- Spec: `docs/superpowers/specs/2026-06-07-mvf-vaultgroup-routing-mvp-design.md`
- Plan: `docs/superpowers/plans/2026-06-07-mvf-vaultgroup-routing.md`
- Subsystem doc: `docs/subsystems/vault-group.md` (catalog vLannaAi/noy-db#24)

## Test Plan

- [x] `pnpm --filter @noy-db/hub exec tsc --noEmit` — clean
- [x] `pnpm --filter @noy-db/hub test` — 216 files / 2146 tests pass (incl. 16 federation unit tests)
- [x] `pnpm --filter @noy-db/showcases exec vitest run src/98-vault-group-federation.showcase.test.ts` — 5/5 pass against the built package
- [x] `node scripts/validate-features.mjs` — OK (registers `vault-group-federation` + showcase 98)
- [x] `packages/hub/src/query/join.ts` unchanged vs `main`

### Notes for reviewers
- `pnpm --filter @noy-db/hub bundle-check` currently fails on a **stale repo-wide baseline** (last reseeded at vLannaAi/noy-db#157, before snapshots/i18n/cross-join/etc.) — it is not wired into CI and federation is verified as a separate chunk with no core leak. Baseline refresh is a separate release-time task, intentionally not folded into this branch.

### DIFF
--- a/SUBSYSTEMS.md
+++ b/SUBSYSTEMS.md
@@ -21,7 +21,7 @@ const db = await createNoydb({
 
 When a subsystem is not opted into, its real implementation is replaced by a NO-OP stub (or a throwing stub on opt-in surfaces) and the heavy code is fully tree-shaken from the bundle.
 
-This document lists the always-on core and the 23 subsystems. It is the table of contents for the rest of the documentation.
+This document lists the always-on core and the 24 subsystems. It is the table of contents for the rest of the documentation.
 
 ---
 
@@ -43,7 +43,7 @@ Anything outside this floor is a subsystem.
 
 ---
 
-## The 23 subsystems
+## The 24 subsystems
 
 Each subsystem has its own subpath export under `@noy-db/hub/<name>`, a `with<Name>()` factory, and a doc page in `docs/subsystems/<name>.md`. The "LOC saved" column is the bundle weight a consumer avoids by **not** opting in.
 
@@ -114,8 +114,9 @@ The Dim 14 family. All three share the same encrypted-payload metadata envelope,
 | # | Subpath | Headline | LOC saved | Pairs with |
 |---|---|---|---:|---|
 | 17 | `@noy-db/hub/routing` | Multi-store routing + middleware + sync-policy + lazy-mode + LRU cache | ~1,985 | `indexing`, `bundle` |
+| 24 | *(preview)* | Multi-vault partition federation — `db.openVaultGroup()` transparent shard routing + `vault-registry` source-of-truth + `minVersion` fan-out guard (MVP, milestone 16) | — | `queryAcross`, `permissions` |
 
-**Totals:** ~16,650 LOC across all 23 subsystems are tree-shake-able. A consumer using only the core ships ~6,500 LOC. A consumer opting into all 23 ships ~31,700 LOC.
+**Totals:** ~16,650 LOC across all 24 subsystems are tree-shake-able. A consumer using only the core ships ~6,500 LOC. A consumer opting into all 24 ships ~31,700 LOC.
 
 ---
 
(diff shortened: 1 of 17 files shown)


vLannaAi:
**Updated:** rebased onto vLannaAi/noy-db#315 (the vLannaAi/noy-db#313 security fix, which owns the `openVault` `create` flag) and folded in the vLannaAi/klum-db#13 forward-compat amendments. This PR is now **stacked on vLannaAi/noy-db#315** — review/merge vLannaAi/noy-db#315 first.

Adds on top of the VaultGroup routing MVP:
- **A — `'no-grant'` skip reason** + `classifyShardSkip` (`NoAccessError` → `'no-grant'`; `InvalidKeyError`/corruption → `'error'`, never masked).
- **B — key-custody-neutral fan-out**: `toArray` opens shards with `create:false` and classifies skips; `openShard` is non-creating so `shard()` drill-down and `put`-routing fail closed too (`createShard` is the sole creating path). Test asserts a scoped advisor's fan-out/drill-down/put all yield `no-grant`/`NoAccessError` and **write zero keyring**. Contract + opaque-`keyOf`/roster guidance in `docs/subsystems/vault-group.md`.
- **E — optional `Reducer.merge`** (sum/count/avg/min/max) for parallel/hierarchical combine (protocol seam; central-reduce stays default).
- Federation entry exports made **type-only** (chunk stays lazy for CJS/non-tree-shakers too).
- Kernel ceiling `noydb.ts` 2920→2960 (genuine core: vLannaAi/noy-db#313 gate + federation entry points; extractable logic already moved out).

Full suite 2195, typecheck, validate-features, architecture, bundle leak-canaries all green. The cross-vault live/aggregate slice (Phase 3) is **not** here — separate follow-up.

PR #453, opened 2026-06-16
docs: consumer migration guide (federation → @klum-db/lobby)

vLannaAi:
Migration guide for consumers after the federation extraction (#450/#452): the grep test to decide if klum-db is needed, niwat's local-consumption setup (yalc/tarball) + single adapter module + call-site rewrites, and i3speedex's likely no-op path (federation-free → stays on @noy-db/hub). Keyed by app name; documents the breaking change + the don'ts.

### DIFF
--- a/docs/migration/2026-06-16-federation-to-klum-db.md
+++ b/docs/migration/2026-06-16-federation-to-klum-db.md
@@ -1,0 +1,167 @@
+# Consumer migration: federation → `@klum-db/lobby`
+
+**Date:** 2026-06-16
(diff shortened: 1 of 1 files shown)


vLannaAi:
Closing in favor of an anonymized version. This guide named the two private pilot consumers directly (and their domains), which violates the 'never name the private pilot client' rule. A redacted rewrite — same content, consumers referred to only by role (Consumer A = federation user, Consumer B = federation-free) — supersedes it. The names live in this branch's history, so this PR/branch should not be merged or revived.

Issue #613, opened 2026-07-10
Period-driven cold archival — vault.archivePeriod (#604 Spec 3) [api]


vLannaAi:
Spec 3 of the #604 period-close lifecycle arc — the deferred cold-archival half.

**Deliverable:** `vault.archivePeriod(name)` relocates a **closed** period's in-window sealed records from the hot store to the configured cold tier, keyed to the period's `_ts` upper bound (same boundary freeze uses), driving `routeStore`'s existing hot→cold migration + read-through. Non-destructive, idempotent, records a `_period_archives/<name>` companion + ledger entry (parallel to freeze). Summaries (`_`-prefixed) stay hot.

**Reframing (from brainstorm):** most cold-tier machinery already exists on the `NoydbStore` seam — `routeStore({ age: { cold } })` does the hot/cold split + read-through, `compact()` migrates, `withLazy()` gives true memory-cold. The only gap: migration keys off a *rolling* `coldAfterDays`, decoupled from periods. This spec adds an **explicit period-derived cutoff**: `compact(vault, { before })` + a `Vault._archiveClosedPeriod(before)` seam + the `archivePeriod` operator call.

**Surface:** `api` — rides the existing `NoydbStore` seam; no `/adapter` change. Store must be a `routeStore` with a cold route (capability-gated; clear error otherwise).

**Key decisions:** `_ts` boundary not business-date (store can't see encrypted business dates; late-`_ts` records archive next period — same as late-booked deletes); separate call from `freezePeriod` (archive is non-destructive, gated only on `closed`); hydrated mode is storage-cold not RAM-cold (`withLazy()` for memory-cold). Out of scope: #611 cross-target purge (different axis), un-archive, business-date precision.

Design doc: `docs/superpowers/specs/2026-07-10-period-cold-archival-design.md`.

Issue #604, opened 2026-07-09
feat: period-close lifecycle + operator purge/archive (Spec 2 — consumes #589 delete markers)


vLannaAi:
**Spec 2 of the #589 arc.** #589 (Spec 1) adds version-ordered delete markers that converge on pull and are **never auto-GC'd** — provably no resurrection, but storage grows monotonically with distinct-deleted ids. This issue is the deliberate, operator-gated cleanup that reclaims that space, framed as a period-close lifecycle.

## Concept

A **period** moves through states:

1. **Open** — normal operation, records mutate freely.
2. **Closed** — no more changes accepted; an end-of-period **summary** is produced. Still **re-openable** (a close is reversible until it's frozen).
3. **Frozen / purged** — triggered manually or on a schedule some time after close. Delete markers in the period are **physically purged**; live records are **compressed and optimized for cold usage**; **only the summaries stay hot**. Effectively re-starts the collection's active set around the summaries + surviving live data.

A cutoff can be expressed as a date (`before X`) or a duration back from now (`older than D`).

## Why operator-gated, not automatic

The sync engine has **no peer/watermark/acknowledgement state** (verified in the #589 exploration) — in a serverless zero-knowledge system nothing can compute "every peer has seen this delete," so automatic GC-once-converged is not available. The period-close model replaces that impossible-to-compute safe-point with a **domain-meaningful, human-asserted one**: "this period is closed and reconciled" is a declaration that all participants have settled.

## Load-bearing safety invariant

Purging a delete marker **re-opens the #589 resurrection window** for any peer that was offline since before the cutoff and never observed the marker. Never-GC is safe *because* the marker is always present to win convergence. Therefore the purge must be an explicit, logged, operator-asserted safe-point — which is exactly what a closed-and-settled period provides. The period lifecycle is not a convenience wrapper; it is what makes purging **safe**.

## Seam it builds on (from #589)

- `vault._purgeDeleteMarkers(before: ISOTimestamp, collections?)` -> count — physical removal of delete markers older than a cutoff, ledger/event-logged. Spec 1 ships this minimal primitive + tests; this issue builds the lifecycle on top.

## Scope to design (own brainstorm)

- Period state model (open/closed/frozen) + persistence; re-open semantics.
- Summary production + preservation (what a summary is; where it lives; must survive freeze).
[...]

Issue #589, opened 2026-07-07
Sync carries no delete tombstones — deletes never converge on pull; offline peers can resurrect deleted records


vLannaAi:
## Verified behavior (`with-party/team/sync.ts`)

- `pull()` is snapshot-based (`remote.loadAll`) and **additive-only**: it puts new/newer remote envelopes locally but never deletes a local row. A record deleted remotely simply stops appearing in the snapshot; peers that hold a local copy keep it forever.
- Deletes propagate **push-side only**, as bare `remote.delete(vault, collection, id)` — no version check, no marker left behind.
- Consequence: after the remote envelope is gone there is **no durable evidence a delete ever happened**. "Deleted" is indistinguishable from "never synced here". An offline peer that edits its surviving copy will push the record back — silent resurrection, for **any** collection.

## Why it surfaced now

The satellite-collections design (`docs/superpowers/specs/2026-07-05-satellite-collections-design.md`, § Convergence & existence authority) needed a cross-collection delete-convergence rule and found no delete evidence to build on. The feature **contains** the gap at the read layer (base row = sole existence authority; a base-less satellite is dead ciphertext, unreachable through every handle) rather than fixing it — and leans on `forget()`'s persistent in-place tombstone as the one durable delete-evidence marker that exists.

## General fix direction

A sync-level delete tombstone / oplog (so `pull()` can observe and apply deletions, with a retention/compaction story), owned by the sync subsystem — not by any single feature. Interacts with: conflict resolvers (delete-vs-edit conflicts are currently unrepresentable on pull), `modifiedSince` partial sync, and the satellite pairing sweep (a real tombstone would allow physical GC of dead-ciphertext satellites, currently retained by design).

## Exit criteria

Revisit when (a) delete convergence produces a real-world resurrection report, (b) the satellite dead-ciphertext retention becomes a storage-cost complaint, or (c) the sync engine gets its next structural iteration.

vLannaAi:
Notes from the #590 fix's final review (PR #602), for this issue's design:

1. **Dirty-delete removes shred evidence**: a peer with a local dirty `delete` for a record that was crypto-shredded elsewhere will pull the tombstone (as a new envelope), then push `remote.delete` — removing the durable erasure evidence from the remote. No plaintext returns, but a third live-holder can then re-push its copy with no CAS obstacle. A real delete-tombstone design should treat shred tombstones as undeletable evidence.
[...]

PR #821, opened 2026-07-25
feat(hub): period-scoped sync pull — thin-client bootstrap (#807)

### LINKED ISSUE #807: sync: period-scoped pull — thin-client bootstrap (current period first, history on demand)
Part of milestone [LINE/LIFF client portal](https://github.com/vLannaAi/noy-db/milestone/36).

## Summary

**Period-scoped sync pull** — thin-client bootstrap: a fresh device pulls the *current period* first and fetches historical periods on demand, bounding first-sync download on 4G.

## Context — what already exists (no klum-db required)

Period partitioning for a **single vault is pure hub**: `archivePeriod` relocates closed-period records to a cold `routeStore` route, and with `withLazy()` cold records are fetched only on access. klum-db is only involved for *fleet-wide* period operations (cross-vault purge coordination) — not needed here. What's missing is the **sync dimension**: `PushOptions`/`PullOptions` filter by collection today, not by period, so a fresh device's first pull still downloads all hot records of a collection regardless of period. `fr7-surface-scoped-sync` documents the direction.

## Scope

- `PullOptions.periods?: string[] | { current: true }` (exact shape in-spec): first sync pulls current-period records + period summaries; archived/older periods excluded.
- On-demand backfill: `pullPeriod(periodId)` (or an option on pull) fetches one historical period into the local store — triggered by UI navigation (deep link into an old period must trigger it transparently).
- Interaction rules to spec carefully:
  - period summaries stay hot and always sync (they're the navigation index);
  - delete-marker convergence per period (freeze/purge machinery) must tolerate a device that never pulled that period;
  - dirty-queue push is NEVER period-filtered (client writes always flow up).
- KPI instrumentation hook: expose bytes/records pulled per phase so the showcase can demonstrate the 4G budget.

## Non-goals

- Fleet/cross-vault period ops (klum-db).
- Byte-level partial records; granularity is the period.

## Tasks

- [ ] Spec the option shape + marker-convergence rules against `fr7-surface-scoped-sync`
- [ ] Implement pull filter + on-demand period backfill
- [ ] Tests: fresh-device bootstrap size, backfill idempotence, marker convergence with never-pulled periods, push unaffected
- [ ] Docs: thin-client bootstrap section in periods + sync subsystem pages

vLannaAi:
Closes #807. Milestone: LINE/LIFF client portal [api].

## What

Thin-client bootstrap for the portal: a fresh device pulls the **current period only** and backfills history on demand.

- `PullOptions.periods?: string[] | { current: true }` — matches the existing `collections`/`modifiedSince` option idiom; backfill = re-run `pull({ periods: [...] })`, idempotent; no new engine methods on the barrel.
- **Membership rule** (the design decision): envelope write-time `_ts` against closed periods' exclusive upper bounds — the established store-tier law (freeze #604 / archive #613); the `dateField` business-date seal is unreadable in ciphertext, so late-booked records ride the current period, mirroring archive. Closed periods partition the timeline; `{ current: true }` = at-or-after the latest closed bound (= everything when no periods are closed — nothing silently dropped).
- **Never filtered**: `_periods` + companion summaries (phase 1 — the navigation index), delete markers/tombstones (#589/#590 law — proven by three convergence tests incl. a frozen-away-markers remote), reserved lookups.
- **Push is never period-filtered** — no `periods` member exists on push options at the type level.
- `PullResult.phases` — per-phase `{ records, bytes }` counts, the 4G-budget KPI hook the showcase (noy-db-docs#120) will chart.
- Guard: a period-scoped pull against a remote whose `_periods` the client can't read throws with a `withPeriods()` hint instead of silently full-pulling.

## Verification

20 new tests (bootstrap, backfill idempotence, middle-window bounds, deep-link late access, filter intersection, push-unfiltered, 3× marker convergence, real `closePeriod()` E2E, 6 validation cases). Full hub suite **526 files / 4457 tests pass** · typecheck · lint · build · `check:architecture` ✓ (noydb.ts ceiling 2419→2420, one wiring line; enclave golden +`envelopeBodySize` per its baseline procedure) · knip no new findings. Changeset: `@noy-db/hub` minor.

## Flagged follow-up (pre-existing, filed separately)

`closePeriod` writes via `writeReserved`, which never dirty-tracks — `_periods` summaries don't push from an authority device with a private local store. Filed as its own issue.


vLannaAi:
CI Build failure was the bundle-size gate: all-on +9.4% (9,644→10,546 gz) — over the 5% per-change tolerance. Investigated before accepting: every scoped scenario is flat (±0.5%), the floor is untouched, all leak checks pass, and the delta (~900 gz ≈ ~350 source lines minified) matches the feature's own weight in the sync chunk, which has no individual scenario and therefore only surfaces in all-on. Baseline consciously updated in `bundle-manifest.json` (the updater script refuses while any failure exists — including the one being accepted — so this is the equivalent reviewable hand-edit, same pattern as the surface goldens).

PR #833, opened 2026-07-26
feat(hub): complete the /cargo + /pod surface ahead of /bundle retirement (#812, #820)

vLannaAi:
Part of #812 — the step-1 completion the klum 0.4.0-pre.1 evaluation surfaced. **Additive only; nothing is retired here.**

## Why

The pre.1 promotion was function-complete but not surface-complete. Re-checking before cutting `/bundle` found **22 exports that would be orphaned** (no published home). This closes that gap:

| Group | Lands on | Why |
|---|---|---|
| `WalkClosureOptions`, `ClosureResult`, `DanglingRefNotice`, `ExtractPartitionResult` | `/cargo` | The promoted functions' **own** option/result types — a caller could call `walkClosure()` but not name its options. A defect in what shipped in pre.1. |
| `adoptPartition`, `unsealDeks`, `createOwnerOnAdoptedPartition` + 6 types | `/cargo` | The **adopt half** of the transfer ceremony; extraction without adoption is half a feature (`transferable-partitions.md` documents the pair). |
| `TransferSealError`, `AdoptionStateError`, `PartitionExtractionError` | `/cargo` | Thrown by the transfer ops (split by actual throw site). |
| `BundleIntegrityError`, `BundleSealMismatchError`, `PodVersionConflictError`, `BundleVersionConflictError`, `BackupLedgerError`, `BackupCorruptedError` | `/pod` | Thrown by the artifact/backup paths — `instanceof` from the subpath instead of the root barrel. |

The three internal alias tests now assert the **canonical** seam first and keep their `/bundle` assertion explicitly marked for deletion alongside the subpath.

## After this

`/bundle` has no export lacking another home, so the remaining retirement steps are: noy-db-docs swaps 2 showcases + 3 doc pages to `/pod` (can proceed against pre.1 today), then `./bundle` + `src/legacy/` are deleted in a **later** release — deliberately not the same one, so a window exists where both old and new resolve.

## Verification

Full hub **529 files / 4558 tests** · typecheck (incl. the extended compile-time `_FrozenTypes` list) · lint · build · `check:architecture` ✓ · bundle-check ✓ (re-exports of already-bundled modules — zero movement) · cargo golden values 30→36 / types 34→44, pod golden values 24→30, both consciously baselined. Changeset: `@noy-db/hub` minor.


vLannaAi:
Folded in **#820** (`hasNoydbBundleMagic` → `/pod`) — same pre-retirement surface-completion work, flagged back from klum-db#45: the predicate sat beside `NOYDB_BUNDLE_MAGIC` everywhere except the subpath, forcing an otherwise-migrated consumer to keep one root-barrel import alive. Pod golden 30→31.

